PR #1, opened 2020-08-12
Proposed repo structure

jsignell:


jameslamb (approved):
Yeah I love this! Thanks for the `tree` example in the README, that's exactly how I imagined it

PR #2, opened 2020-08-17
Add repo structure validation

jameslamb:
~This PR is still a work in progress. Its goal is to propose a repo structure consistent with what we agreed to in #1 .~

UPDATE (8/18/2020): some of the original changes in this PR have been broken out into others: #3, #4, #5, #6 . This PR's focus has been narrowed to just adding the code to validate examples in the `examples/` directory.

It introduces some code to validate examples on every build in CI, to make it a bit friendlier to develop under the guidelines for this repo. This description will be updated once the PR is ready for review.


jameslamb:
can confirm that GitHub Actions is doing what we want!

image

https://github.com/saturncloud/examples/pull/2/checks?check_run_id=999824493

jameslamb:
Ok this PR is ready for review! This proposes a continuous integration job that will validate the structure of example folders, so we can have higher confidence in releases.

.ci/validate-examples.py
```diff
@@ -0,0 +1,205 @@
...
+
+        with open(saturn_json, "r") as f:
+            saturn_config = json.loads(f.read())
+
+        for required_key in SATURN_JSON_KEYS:
+            if required_key not in saturn_config.keys():
+                msg = f"'{saturn_json}' missing required key: '{required_key}'"
+                ERRORS.add(msg)
+
+        project_image = saturn_config["image"]
```
rikturr, line 195: should there be some validations for `jupyter` also?

jameslamb: definitely! Is it ok if I do that in a follow-up PR after this one? I want to take this small steps at a time.

rikturr (approved):

examples/examples-cpu/.saturn/start
```diff
@@ -0,0 +1,10 @@
+sudo apt update
+sudo apt-get install -y htop
+
+pip install \
+    --upgrade \
+        dask \
+        dask-saturn \
+        dask-xgboost==0.1.11 \
+        prefect>0.13.0 \
+        xgboost==0.90
```
jsignell, line 10: Is this just a "for now" start script? We are planning on building an examples-cpu image at some point right?

jsignell (approved):
This looks great! I am all for merging and doing follow ons

examples/examples-cpu/.saturn/start
jameslamb: short answer: yes

long answer: My goal is to get the structure to a place where I can test the changes on the Saturn side, and then to leave it to @rikturr to fill out this repo with the example content he wants to add. If that eventually means removing some or all or this start script and swapping in an image, I'm all for it

jameslamb: To be clear, I still think that the concept of start script should always be supported in this project. But this specific one could probably be replaced by moving stuff into an image

PR #3, opened 2020-08-18
[docs] Add more detailed docs on repo structure

jameslamb:
As I was working on #2, I realized I was doing too many things at once. Trying to break it up so the PRs are easier to review.

This PR proposes filling in more details on the repo structure, following up on #1 .

1. adds a `CONTRIBUTING.md` with details on the expected structure
2. makes it clear that this repo will not be used to build custom images, and that that should be done in https://github.com/saturncloud/images

Once we agree on this and it's merged, #2 will just be about moving files around to make what we have there compliant, and adding a continuous integration job that will check that all examples follow this structure.


rikturr (approved):
love the `.saturn` folder

jameslamb:
> love the `.saturn` folder

thanks! I thought the `.` was a good convention to take from CI services (`.github/`, `.circleci/`) and editors (`.vscode/`, `.idea/`)

rikturr:
> > love the `.saturn` folder
> 
> thanks! I thought the `.` was a good convention to take from CI services (`.github/`, `.circleci/`) and editors (`.vscode/`, `.idea/`)

also added benefit that it doesn't show up in JupyterLab file browser

CONTRIBUTING.md
```diff
@@ -0,0 +1,89 @@
...
+    - `disk_space`: An amount of disk space, in units `Gi`. See the "Disk Space" dropdown on the "Jupyter" page in the Saturn UI for valid values.
+    - `ssh_enabled`: A boolean indicating whether to set up [SSH access from outside Saturn into the Server](https://www.saturncloud.io/docs/connecting/tools/ssh/)
+* `environment_variables`:
+    - A dictionary whose keys are the names of environment variables, and whose values are the values for the environment variables.
+
+Any customization of Dask clusters should be done in notebook code, using [`dask-saturn`](https://github.com/saturncloud/dask-saturn).
+
+#### `.saturn/start` <a name="start-script"></a>
+
+A shell script that will be run on startup of the Jupyter server and all Dask resources. This should contain code that runs quickly, like `pip install`-ing libraries.
```
jsignell, line 89: Should this maybe just be included in the json? I don't have a strong preference either way.

CONTRIBUTING.md
```diff
@@ -0,0 +1,89 @@
...
+* [`.saturn/saturn.json`](#saturn-json)
+* [`.saturn/start`](#start-script)
+
+#### `.saturn/saturn.json` <a name="saturn-json"></a>
+
+A JSON with the following structure:
+
+```json
+{
+    "image": "saturncloud/saturn:2020.07.08.1",
```
jsignell, line 65: The project name is just the name of the directory right?

README.md
```diff
@@ -1,5 +1,6 @@
 # Examples
-Public facing examples for populating Saturn Cloud installations
+
+Public facing examples for populating Saturn Cloud installations.
 
 ## Structure
```
jsignell, line 5: I think we should just reference contributing from here instead of duplicating content

jsignell (commented):
My main comment is to try to avoid duplication in CONTRIBUTING and README

CONTRIBUTING.md
jameslamb: What would the advantage of that be?

I think that the start script should be a script that looks exactly like what you would put in the `Start Script` box in the UI. Otherwise there will be extra mental effort to figure out how the things in the repo relate to the things you see on the "Jupyter" page.

README.md
jameslamb: ok can do. Let's wait until #6 is agreed and then I'll merge that into this

CONTRIBUTING.md
jameslamb: sure, makes sense

jsignell: this was agreed in #6

CONTRIBUTING.md
jsignell: The same could be said of environment_variables right?

jameslamb: it could. The difference with environment variables is that they'll always be exactly one thing: a set of string keys and string values. The order they're defined is also irrelevant, so they don't depend on each other.

Those things aren't true for the `start` script. The `start` script can be any code, and I think that we'd inevitably end up stitching together some parts of it from `saturn.json` plus some parts that are a plain script if we added a `pip_dependencies` section in `saturn.json`.

jsignell: fair enough :)

jsignell (approved):
Looks great!

jameslamb:
https://github.com/saturncloud/examples/pull/3/commits/1c8cb87aae73a3b06b9b3edc69152ca96ebdb011 removes the duplicated stuff from `README.md` and pushes it to CONTRIBUTING. It also adds a line explaining that directory names under `examples/` will be project names.

PR #7, opened 2020-08-20
add CODEOWNERS

jameslamb:
This PR proposes adding a CODEOWNERS file, so that every pull request will automatically be assigned a reviewer.

This file also allows us to handle ownership of different sections without anyone needing to "just know" things like "@rikturr did the nyc-taxi stuff".

I definitely believe that @rikturr should be on every PR here. @jsignell I assumed you'd want to be, but let me know if you want me to take you off the catch-all rule.


rikturr (approved):

jsignell (approved):
LGTM

PR #8, opened 2020-08-20
add validation on saturn.json

jameslamb:
This PR adds validation on the contents of `saturn.json`. Using `marshmallow`, the validation code should now catch all of these:

* required field is missing
* required field is present, but has wrong type
* required field is present, but null
* unknown extra field added

I intentionally introduced all these problems into `examples-cpu/.saturn/saturn.json` just so we can confirm that the GitHub Actions job fails and gives a meaningful error. I'll revert those changes before merging.

### DIFF
--- a/.ci/validate-examples.py
+++ b/.ci/validate-examples.py
@@ -15,6 +15,8 @@ import re
 import requests
 import sys
 
+from marshmallow import fields, Schema, ValidationError
+
 parser = argparse.ArgumentParser()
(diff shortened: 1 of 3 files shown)


jameslamb:
I don't think we should validate that `disk_space` and `size` are recognized values. That would mean hard-coding some of Saturn's service logic in this repo, and the two repos could drift apart as that set of options changes.

As long as that information isn't publicly accessible, I think we should catch issues like "7xlarge is not a valid size" with internal integration tests.

rikturr (approved):

PR #9, opened 2020-08-20
[docs] proposed strategy for managing releases

jameslamb:
This PR proposes a strategy for releases. See the diff for more details, but my basic proposal is this:

> cut long-lived release branches, and write that branch name into `settings` in `saturn`, so that one release of Saturn always corresponds with one release of this project

This wouldn't mean that users on old versions of `saturn` can't get new examples. Since this is a public repo, they can always just clone this repo in a start script, and we can document how to do that. But it would mean that the *default set of examples* will always be the same for a given version of Saturn.

Doing this will give us confidence that merging something to `main` won't break anything running in an older version of Saturn in some user's environment.

I haven't thought through yet how we'll handle the case of a user *upgrading* Saturn versions, but that can come soon after this. I think the use of release branches is a better way to do this than using git tags.

Let me know what you think! If we agree to this approach, I'll set up a branch protection for branches matching the pattern `release-*`


rikturr:
Is this the same way we handle other repos that get released with saturn? Like [images](https://github.com/saturncloud/images)? Since Julia is out might want to get another set of eyes besides me on this

jameslamb:
> Is this the same way we handle other repos that get released with saturn? 

For `images`, we publish tagged artifacts to Docker Hub. Once created, those never change.

For code packaged as libraries, like `prefect-saturn` and `dask-saturn`, we published versioned artifacts to PyPi. Once created, those never change.

For code like `saturn`, we use these long-lived release branches.

No problem if you don't feel comfortable saying :+1: / :-1: , but we shouldn't wait for Julia to get back to keep making progress on this.

@forana or @hhuuggoo could you look at the diff in this PR and give us an opinion? As a reminder, last week we did agree on this for this first version of example projects ([documented in ClickUp](https://app.clickup.com/t/a8kntb)):

> You get examples set up one time when a user is created and they're just like any other projects from that point forward. They will never be updated.

jameslamb:
From discussion in coffee chat today, it sounds like people are ok with the proposal laid out in this PR:

> a release of Saturn has a setting which controls which branch of `saturncloud/examples` to clone example code from.

rikturr (approved):
approving per coffee chat discussion

PR #10, opened 2020-08-20
Start nyc taxi examples

rikturr:
Right now this just has the dashboard examples and the readme's but want to get some review on it now. I changed a few things in the dashboard notebooks from the full demo, mostly to sample data and use a small cluster. 

Let me know if anything isn't clear in the readme, I want to make sure its super seamless for users to get going. I think as part of the release process we can do some testing with a fresh saturn user to make sure all the pieces fit together, but good to catch anything early now

Validating fails because I don't need any environment variables currently for the example. @jameslamb I can modify the `validating-examples.py` but wanted to check with you first.


rikturr:
Added empty environment variables to examples-cpu now all good

jameslamb (changes requested):
Looking pretty good! I left some suggestions. As a general principle, I think these types of "getting started" examples should just be of the form "I am going to show you how to do X. Do this, then that, then this other thing...ok now you have done X". I left comments on a few places where I think the code or docs is a bit too flexible, and where the usability could be improved by making them stricter.

jameslamb (commented):
I'm good with this, pending the unresolved comments being addressed.

jameslamb (approved):
looks good! let's merge it

PR #14, opened 2020-09-08
use cluster_class instead of address for Dask executor

rikturr:
use `DaskExecutor(cluster_class=SaturnCluster)` to match [docs page](https://www.saturncloud.io/docs/deployments/data-pipelines/#multi-node-cluster)

### DIFF
--- a/examples/examples-cpu/prefect/prefect-scheduled-scoring.ipynb
+++ b/examples/examples-cpu/prefect/prefect-scheduled-scoring.ipynb
@@ -326,17 +326,6 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "cluster = SaturnCluster()\n",
-    "cluster.scale(2)\n",
-    "cluster"
-   ]
-  },
-  {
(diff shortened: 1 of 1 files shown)


jameslamb (changes requested):
since we're getting rid of `.scale()`-ing the cluster, can you add an `adapt_kwargs`?

```python
DaskExecutor(
    cluster_class=SaturnCluster,
    adapt_kwargs={"minimum": 2, "maximum": 2},
)
```

I know we now that setting `n_workers: 1` by default but I'm not sure if that is released yet. This will guarantee you don't end up in a situation where the cluster starts up without workers.

rikturr:
> since we're getting rid of `.scale()`-ing the cluster, can you add an `adapt_kwargs`?
> 
> ```python
> DaskExecutor(
>     cluster_class=SaturnCluster,
>     adapt_kwargs={"minimum": 2, "maximum": 2},
> )
> ```
> 
> I know we now that setting `n_workers: 1` by default but I'm not sure if that is released yet. This will guarantee you don't end up in a situation where the cluster starts up without workers.

Done!

jameslamb (approved):

PR #16, opened 2020-09-11
Snowflake examples - NYC taxi

rikturr:
**examples-cpu/snowflake**
Small example showing how to use Snowflake with Pandas and Dask.

**examples-cpu/nyc-taxi-snowflake**
Re-purpose the NYC taxi demo to utilize Snowflake for data ingest and querying. The `data-aggregation.ipynb` notebook is no longer needed because we can do the aggregations on-the-fly inside `dashboard.ipynb` with Snowflake. The ML notebooks then pull data from Snowflake as well. The only net-new file is `load-data.sql` that does the data ingest into Snowflake.

Still need a bit of cleanup and docs for some pieces but wanted to get draft PR out for visibility


hhuuggoo:
Task linked: [CU-bwgjtr re-purpose taxi demo to pull from snowflake](https://app.clickup.com/t/bwgjtr)

jameslamb:
@rikturr are you looking for a review, or not til it's out of draft? If not til it's out of draft, can you remove us as reviewers?

rikturr:
> @rikturr are you looking for a review, or not til it's out of draft? If not til it's out of draft, can you remove us as reviewers?

its ready for review now! So I know for the future, do we typically not assign reviews for draft PRs? Then I would just ping some people in Slack if I want them to look at it?

jameslamb:
I expect that if you assign me for a review, it means "please review this". It's fine to get a review on a draft PR but that's unusual enough that you should explicitly say on here "hey this is still a draft but I'd like a review before I go any further".

examples/examples-cpu/snowflake/load-data.sql
```diff
@@ -0,0 +1,53 @@
...
+  , tolls_amount float                     --   15     17
+  , improvement_surcharge float            --   16     18
+  , total_amount float                     --   17     19
+  , congestion_surcharge float             --   18     
+);
+
+-- Load 2019 Yellow Cab data
+copy into taxi_yellow from (
+select metadata$filename, $1, $2, $3, $4, NULL, NULL, $5, $6, $7, NULL, NULL, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18
+from @taxi_stage/) pattern='.*trip.data/yellow.*2019.*' file_format = (type = csv skip_header = 1);
```
jameslamb: this is really difficult to read. Could you format this and the other calls like this?

```sql
COPY INTO taxi_yellow FROM (
    SELECT
        metadata$filename,
        $1,
        $2,
        $3,
        $4,
        NULL,
        NULL,
        $5,
        $6,
        $7,
        NULL,
        NULL,
        $8,
        $9,
        $10,
        $11,
        $12,
        $13,
        $14,
        $15,
        $16,
        $17,
        $18
    FROM
        @taxi_stage/
)
pattern = '.*trip.data/yellow.*2019.*'
file_format = (type = csv skip_header = 1);
```

examples/examples-cpu/snowflake/load-data.sql
```diff
@@ -0,0 +1,53 @@
+create database nyc_taxi;
+
+create or replace stage taxi_stage url='s3://nyc-tlc/';
+
+list @taxi_stage/  pattern = '.*trip.data/yellow.*';
+
+create or replace table taxi_yellow (
+    csv_filename varchar                   -- 2020   2016
+  , vendorid int                           --    1      1
```
jameslamb: what do all these numbers in the right margin mean?

examples/examples-cpu/snowflake/snowflake-dask.ipynb
```diff
@@ -0,0 +1,388 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import yaml\n",
+    "import snowflake.connector\n",
```
jameslamb, line 40: does the image or start script need to change? I'd be surprised to learn that `snowflake-connector-python` is in our default images, since it's such a heavy dependency

examples/examples-cpu/nyc-taxi-snowflake/README.md
```diff
@@ -0,0 +1,207 @@
...
+- Value: .yml file contents (below)
+
+The .yml file can specify any arguments that can be passed to `snowflake.connector.connect`, such as:
+
+```yaml
+account: ...
+user: ...
+password: ...
+role: ...
+```
```
jameslamb, line 46: can you link to https://docs.snowflake.com/en/user-guide/python-connector-example.html#connecting-to-snowflake, which describes how to set these? Finding the right value for `account` is not obvious.

examples/examples-cpu/nyc-taxi-snowflake/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+You will need to restart the Jupyter server if you add a Credential while its running. The examples utilize a warehouse called `COMPUTE_WH` at size `Medium`, but you can edit the paramemeters of `snowflake.connector.connect` in the notebooks if you want to use a different warehouse.
```
jameslamb:
```suggestion
You will need to restart the Jupyter server if you add a Credential while it's running. The examples utilize a warehouse called `COMPUTE_WH` at size `Medium`, but you can edit the paramemeters of `snowflake.connector.connect` in the notebooks if you want to use a different warehouse.
```

jameslamb (changes requested):
Looks ok to me on the first pass. I need to pull this branch and open up the notebooks...will do that this afternoon.

For now, I left a few comments and suggestions.

I don't see a way to comment on an empty file in GitHub, so adding one more here...please remove `examples/examples-cpu/nyc-taxi-snowflake/.gitkeep`. It's not necessary any more since you have other code in that directory.

examples/examples-cpu/nyc-taxi-snowflake/model-api.py
```diff
@@ -0,0 +1,47 @@
+import s3fs
```
jameslamb: oh sorry, missed this file! I understand why you copied over the `hypterparameter-*` and `xgboost*` notebooks, since some of their code was changed to use snowflake for data loading.

But since this file is identical to https://github.com/saturncloud/examples/blob/3007d8ca5cc9e8a889aee431c72ea747c3860f92/examples/examples-cpu/nyc-taxi/model-api.py, can you please find a way to avoid duplicating it? For example, maybe the command in the deployment can still point to the file in `nyc-taxi/`.

Duplicating code increases the maintenance burden for this repo (if we remember everything) and the risk of copies of that code becoming inconsistent with each other (if we update one place and forget to update others).

examples/examples-cpu/README.md
```diff
@@ -32,8 +34,31 @@ There are examples for GPU-accelerated ML in the `examples-gpu` project.
...
+
 *You will learn:*
 - How to set up a Prefect flow for scheduled scoring of a statistical model
 - How to use a Dask cluster to distribute work
 - How to deploy the flow and run it on a schedule
 - How to deploy a flow using Prefect Cloud
+
+## Snowflake
+
+This example is in the [`snowflake`](snowflake) folder.
```
jsignell: I prefer no backticks on these folder names since they are not code.

examples/examples-cpu/README.md
```diff
@@ -32,8 +34,31 @@ There are examples for GPU-accelerated ML in the `examples-gpu` project.
...
+*You will learn:*
+- How to query Snowflake and load data into a Pandas dataframe
+- How to query Snowflake and load data into a Dask dataframe
+- How to write to Snowflake with Pandas and Dask
+
+## NYC Taxi using Snowflake
+
+This example is in the [`nyc-taxi-snowflake`](nyc-taxi-snowflake) folder.
+
+The same end-to-end data science pipeline in the "NYC Taxi" example, except using [Snowflake](https://www.snowflake.com/) for data ingest and exploratory analysis. The ML examples then pull data from a Snowflake database rather than from S3.
```
jsignell:
```suggestion
The same end-to-end data science pipeline in the "NYC Taxi" example, except using [Snowflake](https://www.snowflake.com/) for data ingest and exploratory analysis. The ML examples pull data from a Snowflake database rather than from S3.
```

examples/examples-cpu/nyc-taxi-snowflake/README.md
```diff
@@ -0,0 +1,207 @@
+|<img src="img/taxi.png" width="200" /> | <img src="img/saturn.png" width="400" />|<img src="img/snowflake.png" width="400" /> |
+| -- | -- | -- |
+
+# NYC Taxi analysis with Saturn Cloud and Snowflake
+
+The notebooks in this example showcase a data science workflow with NYC taxi data, executed on [Saturn Cloud](https://www.saturncloud.io/) with a database hosted in [Snowflake](https://www.snowflake.com/). For more information on how to connect to Snowflake using Saturn, Dask, and Pandas, refer to the `snowflake` example.
```
jsignell: same comment as above about backticks.

examples/examples-cpu/snowflake/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Query data with Pandas
+
+If your table or query result fits into the memory of your Jupyter client, you can load data into a pandas dataframe using `fetch_pandas_all()` or `fetch_pandas_batch()`.
+
+See [`snowflake-pandas.ipynb`](snowflake-pandas.ipynb).
+
+# Query data with Dask
+
+If your table or query result _don't_ into the memory of your Jupyter client, you can use Dask! Then you can take advantage of using a Dask cluster with Saturn to speed up your computations.
```
jsignell:
```suggestion
If your table or query result _doesn't_ fit into the memory of your Jupyter client, you can use Dask! Then you can take advantage of using a Dask cluster with Saturn to speed up your computations.
```

examples/examples-cpu/snowflake/snowflake-dask.ipynb
jsignell: `snowflake-connector-python` _is_ in our default image: [here](https://github.com/saturncloud/images/blob/5afc87a15da762c82c6a7c78ade38b9cf7a04aa8/saturn/environment.yml#L44). But I think there are missing dependencies in general. For instance none of the dashboard dependencies are in the default image or installed in the config file for this project.

jameslamb:
when you come back to update this, can you also please give the pull request a more informative title? That makes the list of closed pull requests readable as a changelog

image

and gives a small hint when you see the message from the most recent commit in the file view

image

examples/examples-cpu/snowflake/snowflake-dask.ipynb
rikturr: created tasks to update images will do today

rikturr: https://github.com/saturncloud/images/pull/65

rikturr:
Ok ya'll sorry for the delay - this is ready for review again. I also added snowflake GPU examples (thanks @hhuuggoo )

jameslamb (changes requested):
left a few tiny style recommendations, sorry. I'm  cloning this right now so I can review the notebooks in jupyter lab

examples/examples-cpu/snowflake/snowflake-dask.ipynb
```diff
@@ -0,0 +1,388 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "cur = conn.cursor().execute(query, '2019-01-01')\n",
+    "df = cur.fetch_pandas_all()\n",
+    "len(df), df.memory_usage().sum() / 1e6  # memory size in MB"
```
jameslamb, line 92: you'd have to already understand what these functions do to understand the two numbers printed here. Could you make this a little more user-friendly?

```python
print(f"number of rows: {len(def)}")
print(f"size in memory (MB): {df.memory_usage().sum() / 1e6}")
```

examples/examples-cpu/snowflake/snowflake-dask.ipynb
```diff
@@ -0,0 +1,388 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from dask.distributed import Client, wait\n",
+    "from dask_saturn import SaturnCluster\n",
+    "import time\n",
```
jameslamb, line 110: do you have a strong preference for sprinkling the imports throughout the notebook, instead of collecting them at the top?

I like having all the imports at the top, because it means that you find out right away if you have any import issues. Leaving them through the code like this means you could get a few minutes into the notebook and then realize you have to go install something, which means you might have to start over.

I'd prefer to keep them all at the top, but if you disagree it's not worth blocking the PR over.

examples/examples-cpu/snowflake/snowflake-dask.ipynb
```diff
@@ -0,0 +1,388 @@
...
+   "source": [
+    "len(ddf)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<br>\n",
+    "The cell below will execute the Snowflake queries across the cluster and compute the length and size of each partition"
```
jameslamb: this language was a little  confusing to me. The code below it prints the number of rows and the memory usage of *the entire dataframe*, not the per-partition number of rows and memory footprint. I think this sentence could be confusing and make a beginner think that all partitions are the same size.

I also think you should use the language "number of rows" instead of "length". It's a convenience that `len(df)` prints the number of rows, but I personally think talking about the "length" of a data frame is kind of unintuitive, and that "number of rows" and "number of columns" is clearer.

examples/examples-cpu/snowflake/snowflake-write.ipynb
```diff
@@ -0,0 +1,32 @@
+{
```
jameslamb: this notebook is empty...is that intentional? If you're not adding code for it in this PR, can you please remove it?

jameslamb (changes requested):
Ok I pulled this branch and clicked through the notebooks. I have a few additional suggested changes.

examples/examples-cpu/snowflake/snowflake-dask.ipynb
rikturr: At least for now I'd like to keep it as-is. I do to have them spread throughout the examples because they provide context for the different classes/functions that are used. The import issues are slightly mitigated because these are paired with images already that should have anything. Also, the sprinkling is how its done in most of the other examples so I would rather at least keep the same pattern for now, and if we want to change for future we can go through and make everything consistent.

examples/examples-cpu/snowflake/snowflake-dask.ipynb
rikturr: ah yes, will definitely update. Throughout the rest of the examples there's a print in a one-liner and I'll change it to that for consistency

examples/examples-cpu/snowflake/snowflake-write.ipynb
rikturr: its a placeholder, but np I'll remove it and the TBD from the readme

examples/examples-cpu/snowflake/snowflake-dask.ipynb
jameslamb: ok

examples/examples-cpu/snowflake/snowflake-write.ipynb
jameslamb: we should not have unused code or TBD documentation here. This repo is intended to be documentation for Saturn, and having stuff that you have to "just know" to ignore is a distracting, and possibly confusing, experience

jameslamb (approved):
Look ok to me!

In the future, it would be good to break things like this into smaller PRs. The changes to non-snowflake stuff like "change these column names because the column names have changed in the data" were unrelated, and could have been merged earlier. That also would have made this easier to review.

PR #17, opened 2020-09-21
remove prefect install from examples-cpu start script

rikturr:
`prefect` and `prefect-saturn` will be part of the default image for now (https://github.com/saturncloud/images/pull/70)

### DIFF
--- a/examples/examples-cpu/.saturn/start
+++ b/examples/examples-cpu/.saturn/start
@@ -1,4 +1,0 @@
-pip install \
-    --upgrade \
-        "prefect>0.13.0" \
-        "prefect-saturn>=0.1.0"


examples/examples-cpu/.saturn/start
```diff
@@ -1,4 +0,0 @@
-pip install \
```
jameslamb, line 1: since `start` isn't required, can you please just remove this file instead of leaving a blank one here?

jameslamb (changes requested):

examples/examples-cpu/.saturn/start
rikturr: sure thing!

jameslamb (approved):

jsignell (changes requested):
The .saturn/saturn.json needs to be set to the new image as well right? And what is the release process like for these examples? Does the released version of saturn pull from a release branch over in this repo?

forana:
> The .saturn/saturn.json needs to be set to the new image as well right? And what is the release process like for these examples? Does the released version of saturn pull from a release branch over in this repo?

Saturn is configured to pull the head of a named branch. For the 08.28 release, that's set to `release-2020.08.28`.

Upcoming release tag for the new CPU image is here: https://github.com/saturncloud/release-images/blob/release-2020.08.28/config.py#L84

rikturr:
according to https://github.com/saturncloud/examples/blob/main/CONTRIBUTING.md#releasing

we need to create a release branch and also update the saturn.json to point to images. So the release branch exists but the image hasn't been pointed over. I'm going to just close this PR then open another PR on the release branch

PR #19, opened 2020-09-24
Add GH action release PR to main 

jsignell:
Might be a good idea to add this action at the org level actually. But this'll do for now.


jameslamb (approved):

forana (approved):

jsignell:
I'm going to let this sit for a while and see if the top-level thing works if that's ok.

EDIT nevermind turns out the top-level thing isn't supported https://github.community/t/putting-workflow-definitions-in-github-repo/16611/2

PR #18, opened 2020-09-24
Remove start script from examples-cpu and update images

rikturr:
Follow up from #17 

- Remove start script from examples-cpu
- Update images for examples-cpu and examples-gpu to `2020.08.28`
- Also updated descriptions to indicate that there are multiple examples in each project

PR #20, opened 2020-09-24
always run CI everywhere

jameslamb:
Noticed on #18 that CI jobs aren't running on PRs into release branches. When this repo was created, we didn't have a notion of "release branches", and I set the CI up to only run on commits to `main`.

This change will just make it run anytime any code is pushed anywhere in the repo.

### DIFF
--- a/.github/workflows/validate-projects.yml
+++ b/.github/workflows/validate-projects.yml
@@ -2,13 +2,7 @@ name: Continuous Integration
 
 # only run this workflow on new commits to main
 # or PRs into main
-on:
-  push:
-    branches:
-    - main
-  pull_request:
-    branches:
-    - main
+on: [push]
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


rikturr (approved):

.github/workflows/validate-projects.yml
```diff
@@ -2,13 +2,7 @@ name: Continuous Integration
...
 # only run this workflow on new commits to main
 # or PRs into main
-on:
-  push:
-    branches:
-    - main
-  pull_request:
-    branches:
-    - main
+on: [push]
```
jsignell: Need to update comment and I don't think push needs the brackets.

jameslamb: done

jsignell (approved):
:tada:

PR #29, opened 2020-10-20
remove unnecessary imports in Prefect Cloud notebook

jameslamb:
There are a few things imported in the Prefect Cloud example notebook that are unnecessary. They're all Dask-y things that are bundled in `prefect_saturn.PrefectCloudIntegration` and don't need to be separately imported.

This PR proposes removing them.


skirmer:
We don't need to import these things? 🤯  Big if true, I'm on board!

skirmer (approved):

PR #30, opened 2020-10-28
Convert prefect notebook to one-shot

bhperry:
Removing the scheduling components from the non-cloud prefect example. This will allow it to be used for creating a scheduled deployment instead, and for integration-tests as a one-shot.

### DIFF
--- a/examples/examples-cpu/prefect/prefect-scoring.ipynb
+++ b/examples/examples-cpu/prefect/prefect-scoring.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


jameslamb:
@bhperry and I talked about this in DMs but I want to add here for context....ease of integration tests isn't the only reason this is a good change

This example was written before "Saturn scheduled deployments" existed. Now that they do, that's a cheaper and more reliable way to deploy a scheduled prefect flow in Saturn than using a persistent deployment + `flow.run()` with a Prefect schedule. So this change also means our docs will encourage a pattern that customers will get more value from.

jameslamb (approved):
thanks!

PR #40, opened 2020-11-09
fix indenting, typos

jameslamb:
This PR proposes two minor changes.

1. Fixing some small misspellings
2. Using clearer indentation for `SaturnCluster` calls

The choice of parameters to `SaturnCluster` is a really important part of understanding how to work with Dask in Saturn, and I think we should be doing one argument per line so it's easier to see those arguments and their values.

**before**

image

**after**

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,4 +4,4 @@ Example projects for getting started with Saturn Cloud.
 
 ## Structure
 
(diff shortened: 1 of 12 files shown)


skirmer (approved):
No objection from me!

PR #44, opened 2020-11-18
remove unused imports from notebooks

jameslamb:
This PR proposes removing unused imports in the example notebooks.

Unused imports make these notebooks worse in the following ways:

* increase memory footprint
* increase load time
* distract and confuse users
* introduce unnecessary risk of notebook breaking if an import fails (like if `from thing import whatever` fails because `whatever` is removed from `thing`)

I couldn't think of a good way to easily check for these in CI. We could do something gross where all the notebooks are converted to `.py` files and then checked with `pylint` or `flake8` but I think that could take a lot of effort. So for now, this PR just cleans some things up and doesn't introduce tests to prevent them from creeping back in in the future.


examples/examples-cpu/nyc-taxi/hyperparameter-scikit.ipynb
```diff
@@ -54,18 +54,13 @@
...
     "import pandas as pd\n",
-    "import s3fs\n",
     "import warnings\n",
     "warnings.simplefilter(\"ignore\")\n",
     "\n",
-    "s3 = s3fs.S3FileSystem(anon=True)\n",
-    "\n",
     "taxi = pd.read_csv(\n",
-    "    s3.open('s3://nyc-tlc/trip data/yellow_tripdata_2019-01.csv', mode='rb'),\n",
+    "    'https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-01.csv',\n",
```
jameslamb, line 61: since these `-scikit` examples are just reading a single CSV file into a pandas dataframe, I don't think we're getting any benefit from using `s3fs`

examples/examples-cpu/prefect/prefect-cloud-scheduled-scoring.ipynb
```diff
@@ -418,9 +418,9 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.6"
+   "version": "3.7.7"
```
jameslamb, line 421: this change is probably from the fact that I wrote this notebook on my laptop, not in a Saturn environment. So I don't think it's a big deal either way, but we should keep it

examples/examples-cpu/prefect/prefect-cloud-scheduled-scoring.ipynb
```diff
@@ -418,9 +418,9 @@
...
   }
  },
  "nbformat": 4,
  "nbformat_minor": 4
-}
\ No newline at end of file
+}
```
jameslamb, line 426: this is not something I did manually. I see these newline / no-newline changes frequently in diffs when I edit notebooks. No idea where they came from. I don't think they're problematic though.

jsignell:
> We could do something gross where all the notebooks are converted to .py files and then checked with pylint or flake8 but I think that could take a lot of effort. 

I imagine the work that @bhperry has done on integration tests is most of the way there. Maybe it'd make sense to add some linting over there.

examples/examples-cpu/nyc-taxi/hyperparameter-scikit.ipynb
jsignell: Agree.

bhperry:
> > We could do something gross where all the notebooks are converted to .py files and then checked with pylint or flake8 but I think that could take a lot of effort.
> 
> I imagine the work that @bhperry has done on integration tests is most of the way there. Maybe it'd make sense to add some linting over there.

Yeah just waiting on the new release builds to get into main before merging in the updates to integration-tests. They are already converting every example notebook to a python file, so we could definitely add linting there.

The only question is do we want to consider a bad lint on the examples as a failure for integration-tests? It's not really the intended purpose of them, but also the examples are a part of the platform so I don't really have a problem with the idea.

examples/examples-cpu/prefect/prefect-cloud-scheduled-scoring.ipynb
bhperry: yeah I think my vs-code setup stripped those out when I was making changes here. I'll see if I can find a setting for that so it doesn't happen again.

jameslamb: I think you should, for all types of files in all settings, not trim off the final trailing newline...

jameslamb:
In my humble opinion, integration tests are way too late to find out about linting problems and it's preferable to do it in this project's on-every-commit CI to prevent problematic code from being merged in.

I think it would be annoying to be running those integration tests on the way to a release, and only then discover some linting issue that you have to come back and fix in the repo.

jsignell:
> I think it would be annoying to be running those integration tests on the way to a release, and only then discover some linting issue that you have to come back and fix in the repo.

100% agree. 

I was imagining the integration tests running on a cron job and warning about linting issues as they chug along. But you are right that doesn't sound very helpful. Probably was should just steal ideas from the integration test repo if/when we want to automate this.

examples/examples-cpu/prefect/prefect-cloud-scheduled-scoring.ipynb
bhperry: Yep, agreed. This is the only file type where that happens, need to dig into the settings to find out why.

jsignell (approved):
Whoops - forgot to hit approve...

examples/examples-cpu/nyc-taxi/hyperparameter-scikit.ipynb
rikturr: s3fs is required for anonymous connections. If a user doesn't have AWS credentials setup, regular `pd.read_csv()` will fail

jameslamb: that's not true.

Here's an example in an environment with no AWS creds and no `s3fs`:

```shell
docker run -it continuumio/miniconda3 /bin/bash

pip install pandas

python
```

```python
import pandas as pd
df = pd.read_csv(
    "https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-01.csv"
)
df.head()
```

image

```python
import s3fs
```

image

jameslamb: notice that in this PR, I also changed the protocol from `s3://` to `https://`. You probably ran into credentials issues before because of the combination of no creds AND using `s3://`. But every file in S3 is available over HTTPS too, and public ones don't require any authentication :grinning:

rikturr: 🤯🤯

rikturr: TIL. I'll make sure to modify other code we have around that uses `s3://`

PR #48, opened 2020-12-09
Single Instance, multi GPU XGBoost Example

mmccarty:


rikturr:
Thanks for contributing this @mmccarty ! I realized that the examples in this repo are set to run on a single-GPU g4dn instance so there would be no benefit in this case from a LocalCUDACluster. We need to keep these examples on small (cheaper) machines to make sure users don't go over spending limits.

We do have another repo for more general notebooks that don't get pre-populated into user accounts: https://github.com/saturncloud/saturn-cloud-examples. Would you be able to open a PR to that repo instead? And add a small note at the top indicating which machine size you used for the Jupyter Server? This notebook would fit well in the `machine_learning` folder.

If we change the size of the machine for pre-packaged examples I'll make sure to put the notebook back here. We can also test using a SaturnCluster for multi-node single-GPU which could work

mmccarty:
Good idea! Will do!

PR #49, opened 2020-12-09
Update saturn images to latest release

bhperry:
Updated daily cluster to use this branch, rerunning integration tests now
https://github.com/saturncloud/release-images/runs/1526847292?check_suite_focus=true


jameslamb (commented):
having these integration tests is SO nice! I'll approve if the tests pass

jameslamb (approved):
thanks for fixing this and for catching the missing imports! I love the integration tests and I'm not afraid to say it.

PR #52, opened 2021-01-04
Add snowflake logo to snowflake notebooks

jameslamb:
This PR proposes adding the Snowflake logo to Snowflake notebooks, so it's a bit easier to visually tell the difference between those and the non-snowflake ones.

**before**

image

**after**

image


skirmer (approved):
Logos look good to me.

PR #53, opened 2021-01-04
consolidate images in one folder per example type

jameslamb:
This PR proposes consolidating image folders into one folder per example type, `examples-cpu/_img` and `examples-gpu/_img`.

Benefits of this:

* smaller repo (from removing duplicate images) = faster clone times = faster new-user setup
* reduced duplication reduces the risk of slightly-inconsistent versions of the same image (like different scikit-learn logos), and allows us to change those images with less files touched, if necessary

I tried to make this new directory `.img/`, so it would be hidden by default in Jupyter Lab, but the `` tag in HTML doesn't like folders starting with dots (like ``).


skirmer (approved):
Looks good to me

PR #54, opened 2021-01-04
update CODEOWNERS

jameslamb:
The CODEOWNERS for this project was originally set up a while ago, and I think it could be adjusted to reflect the current state of responsibilities.

* makes @skirmer and @rikturr the default reviewers for PRs
* makes @jsignell and myself the default reviewers for automation pieces

The CODEOWNERS is just an educated first guess, and we should still all feel comfortable pulling people in as reviewers directly for specific things.


skirmer (approved):
Makes sense to me

jsignell (approved):

PR #56, opened 2021-01-04
remove dependence on other websites for images

jameslamb:
This PR proposes changing all cases where images are being read over the internet to instead use image files checked into the repo. I think this can be a small improvement in the usability of the notebooks. It avoids missing images causes by brief network interruptions, weird browser cache stuff, or removal / movement of the images on the websites we're pulling them from.

I think the improvement in usability is worth the few kB added to the repo by this PR.


rikturr (approved):
🚀

PR #57, opened 2021-01-04
make notebooks the most important thing in READMEs

jameslamb:
I've seen in the user testing we've done so far that when people open up the example project, they get the examples README and see a bunch of directories and then think "hmmmm....what do I do next".

This PR proposes a small change to help with that. It makes the top of the README act like a clickable table of contents for all the notebooks you'll find. When these links are clicked in Jupyter Lab, they'll open up a new notebook:

image

This PR also removes a leftover `saturn.png` and replaces references to it with references to the one in `_img/`.


rikturr (approved):
This is great!

PR #66, opened 2021-01-08
split up snowflake connection and data loading

jameslamb:
Today, the snowflake example notebooks have a section called "Load data and feature engineering". This PR proposes breaking that up into "Connect to Snowflake" and "Load Data". I think that connecting to snowflake is complicated enough that it deserves to be its own section.

This also adds a bit more information to the "Connect to Snowflake" section. We currently ask people to consult the README, but in my current effort I'm trying to make the notebooks more self-contained so that people don't get discouraged or confused as they work through them. This means more duplication but I think it's worth it.


rikturr (approved):
Thanks for this @jameslamb! Very timely as our blog and webinar should be driving more people to these snowflake examples

PR #74, opened 2021-01-11
Update examples readme with getting started instructions

rikturr:
Update readme with getting started instructions. Apologies for the horrendous branch name, I was trying out the clickup integration


hhuuggoo:
Task linked: [DEV-1525 remove "getting started" links from examples because that just confuses people in user testing (since after they get to the notebook, they've already gotten started"](https://app.clickup.com/t/1250339/DEV-1525)

jsignell (changes requested):
Sorry if this wasn't clear, but I think the getting started instructions in examples/*/README.md need to be removed and added to this top level one.

rikturr:
> Sorry if this wasn't clear, but I think the getting started instructions in examples/*/README.md need to be removed and added to this top level one.

Are you referring to this section? This is the only getting started instructions I see in examples/*/README.md at this point

image

The only thing with moving this section to the top-level readme is that users won't see that blurb from within Saturn, so they might not actually know theres a github repo. If that's the goal then I'm okay with it.

edit: I do see one section with the getting started link in the examples/*/README.md files. I'll remove those

jsignell:
> The only thing with moving this section to the top-level readme is that users won't see that blurb from within Saturn, so they might not actually know theres a github repo. If that's the goal then I'm okay with it.

I prefer to move that section up and if we want to make them know that examples are on github, just add a link that says that.

rikturr:
Ok moved the "running in a new project" section to the top-level readme, and added a "next steps" section to each project's readme with some links which include one to this repository

jsignell (approved):
:rocket:

PR #76, opened 2021-01-14
Get rid of links to dirs

jsignell:
Clicking on these links from within jlab leads to error messages like:
image


jameslamb (approved):
thanks!

PR #78, opened 2021-01-20
Feat/quickstarts

hhuuggoo:


jsignell:
Looks like 
image

rikturr (changes requested):
Have a few comments, mostly around the lightgbm notebook. 

- Is there a reason to have it in its own folder? I think it would be better to sit with the rest of the notebooks in `examples-cpu/nyc-taxi`. 
- It looks like the notebook has the format of the old examples before @jameslamb revamped their structure and readability. It would be best to follow the pattern there so as to not confuse users (`examples-gpu/nyc-taxi/02-rf-rapids-dask.ipynb`, edited in this PR, is a good reference). 
- Our default saturn images don't currently have `lightgbm,` so I don't think we want to merge/release this until we get it included in the examples-cpu image

I might recommend splitting the lightgbm part into a different PR so we can get the examples-gpu notebook renaming merged

For the examples-gpu changes, the files are referenced a lot in links so those need to be updated in:
-  examples-gpu and folder readmes
- several references in each notebook

jameslamb:
I'm not sure what the status of this branch is. If it's something we eventually hope to merge to `main`, I think the LightGBM example can be removed. The LightGBM quickstart is now readying from https://github.com/saturncloud/workshop-lightgbm-dask.

hhuuggoo:
> I'm not sure what the status of this branch is. If it's something we eventually hope to merge to `main`, I think the LightGBM example can be removed. The LightGBM quickstart is now readying from https://github.com/saturncloud/workshop-lightgbm-dask.

I'll remove the PR.  this branch is where I'm messing around with stuff still. (hence the draft status)

PR #81, opened 2021-02-05
fix missing arg in feature engineering

jameslamb:
In recent refactoring, `prep_df()` (the function used for feature engineering in these examples) was changed to take in an argument `target_col`, a string with the name of the target column.

Some calls to that function were not updated. This PR fixes that.


bhperry (approved):
LGTM

PR #82, opened 2021-02-05
fix missing arg in feature engineering (#81)

jameslamb:
Ran the following to pick this change from #81 off of `main`.

```shell
git cherry-pick 5fd56fd529ad137493b8fd92bea4863f1b3fcc86
```


bhperry (approved):
LGTM

PR #92, opened 2021-02-12
use shared snowflake credentials for cpu examples

rikturr:
Update documentation and code to use the EXAMPLE_ credentials. Dashboard required pre-creating the view and I'll have a corresponding PR for that in a private repo


examples/examples-cpu/.saturn/saturn.json
```diff
@@ -14,7 +14,13 @@
...
         "MODEL_FILE": "xgboost.pkl",
-        "PREFECT_CLOUD_PROJECT_NAME": "dask-iz-gr8"
+        "PREFECT_CLOUD_PROJECT_NAME": "dask-iz-gr8",
+        "TAXI_DATABASE": "saturn_nyc_taxi"
     },
+    "required_secrets": [
+        "EXAMPLE_SNOWFLAKE_ACCOUNT",
+        "EXAMPLE_SNOWFLAKE_USER",
+        "EXAMPLE_SNOWFLAKE_PASSWORD"
+    ],
```
bhperry: Also requires `PREFECT_USER_TOKEN` for the prefect cloud flow notebook

bhperry (approved):
LGTM

examples/examples-cpu/.saturn/saturn.json
```diff
@@ -14,7 +14,14 @@
...
-        "PREFECT_CLOUD_PROJECT_NAME": "dask-iz-gr8"
+        "PREFECT_CLOUD_PROJECT_NAME": "dask-iz-gr8",
+        "TAXI_DATABASE": "saturn_nyc_taxi"
     },
+    "required_secrets": [
+        "EXAMPLE_SNOWFLAKE_ACCOUNT",
+        "EXAMPLE_SNOWFLAKE_USER",
+        "EXAMPLE_SNOWFLAKE_PASSWORD",
+        "PREFECT_USER_TOKEN"
+    ],
```
jameslamb, line 25: what does this field do? I don't understand this change, sorry.

Every field in these files has special meaning in the code in `saturn` that processes them, and I'm confused by this addition without a corresponding pull request in `saturn`.

jameslamb (changes requested):

examples/examples-cpu/.saturn/saturn.json
bhperry: I am updating integration-test to create projects as part of the test, rather than relying on pre-existing ones. I asked for this field to be added in order to allow it to more intelligently create secrets for the projects that need them, and was thinking that when we move to having secrets explicitly attached to projects we could do something similar on the saturn side.

jameslamb: ok thanks for that context.

If it's just for the integration tests then for now I'd prefer that we put it in a separate file and not touch `saturn.json`. Since `saturn` relies on that file, I don't want to get into a situation where you have to "just know" which fields in there Atlas cares about and which ones it ignores. I think that separation is useful when you're debugging. You can know that changes to `.saturn/required-secrets.json` (or whatever) should have 0 impact on `saturn`.

bhperry: IMO it's useful context to have in saturn, whether or not it's being used directly by saturn. Currently the situation is you have to "just know" that these secrets need to exist, or else the examples will fail. Adding metadata to describe what the project expects to exist is a good thing.

bhperry: I agree that we shouldn't add a ton of stuff to `saturn.json` that saturn doesn't know about, but to me this is a special case that merits it. Not gonna fight over it though if you feel strongly otherwise.

jameslamb: alright fair enough, I don't feel so strongly that it's worth changing. It's a good point that we're already in a "just know" situation with the secrets haha

In the future @rikturr , for a PR like this it would be useful to link to the relevant prior discussion (or summarize it) in the description so reviewers know what they're looking at.

jameslamb (approved):

PR #102, opened 2021-03-10
Examples v2.0 dashboard

jnolis:
This is the first of several PRs to add new quickstarts for Saturn. Since this is the first, it also includes cleaning up of the repo and removing the cpu and gpu examples. Specifically it:

* Has a rewriten readme.md to cover the new structure of each quickstart
* Removes the examples-cpu and examples-gpu folders. The materials in these are being migrated to other locations
* Adds the first dashboard quickstart. This quickstart now:
  * Has an updated `saturn.json` metadata structure.
  * Removes a lot of the "non-dashboard" material (like pulling from s3), to make it more accessible for people who may not understand those technologies (like s3).
  * Moves the all of the material required by the quickstart into the git repo itself (like the csv files of data). This will make it substantially easier to maintain the quickstarts since all the material is in one place. Since the quickstarts are often used as examples and other documents outside of the actual Saturn projects, the code instead of just loading from the directory instead pulls from the github URL. it's a little weird but I think this is the best we can to do balance maintainability and usability across all our channels (examples docs, quickstarts, blogs, etc).
  * Should be our "single canonical example" of how to quickly get a dashboard running in Saturn. We should no longer have a separate "example doc" "examples repo" "saturn-cloud-examples repo" etc etc that we have to maintain, instead it should all just use this one location.


jameslamb (changes requested):
The changes in this PR require corresponding changes to Saturn itself and to Saturn's integration tests.

Before I review I want to know...have you and @jsignell talked about that?

examples/dashboard/dashboard.ipynb
```diff
@@ -0,0 +1,242 @@
+{
```
rikturr, line 1: <p>The thing about reading from github URLs here is that its hard-coded to the branch name, so when the branch is merged/deleted the URL won't work. Isn't this effectively the same as reading from s3?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/saturncloud/examples/pull/102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

rikturr:
Added a comment about Github URLs vs S3. I tend to lean toward using something like S3 because it avoids showing a "bad pattern" of including data in the Jupyter server. For example, if a user uploads a data file into a Jupyter server, but then creates a cluster or deployment(s) that file won't exist on the other machines. If its S3, all machines would have access without any issues.

examples/dashboard/dashboard.ipynb
jnolis: Yeah, I'm not thrilled with that tbh, but I think once this is approved we can hardcode it to `main`.

jnolis:
> Added a comment about Github URLs vs S3. I tend to lean toward using something like S3 because it avoids showing a "bad pattern" of including data in the Jupyter server. For example, if a user uploads a data file into a Jupyter server, but then creates a cluster or deployment(s) that file won't exist on the other machines. If its S3, all machines would have access without any issues.

I think "storing data on the Jupyter server is a bad habit" is something we should think about. Like, so many of our customers will be used to having data stored right with there code, that making this something we have to train customers out of could be a huge struggle. I don't know if it's possible to instead have this be a more permanent data store, but like I want to make sure we don't put our needs (easy to write the backend for) above the needs of customers. That's kind of tangental to the S3 v GitHub thing.

forana:
> Per a conversation with @jsignell, I added back the examples-cpu and examples-gpu since we'll still need them for existing enterprise customers. Those folders were added and a note was added to the readme as well.

Worth noting that there's a branch cut from this repo (`release-YYYY.MM-DD`) cut as part of our standard release process, and that's what's used from customer installs built from that release - it's not _impossible_ to make breaking changes like that, but it has to be coordinated with the release cycle.

examples/dashboard/.saturn/saturn.json
```diff
@@ -0,0 +1,12 @@
+{
+    "image": "saturncloud/saturn:2021.02.22",
+    "jupyter": {
+        "size": "medium",
+        "disk_space": "4Gi",
+        "ssh_enabled": false
+    },
+    "description": "Deploy a dashboard to Saturn Cloud",
+    "title": "Deploy Dashboards",
+    "thumbnail": "thumbnail.png",
```
jsignell: same comment as above about the name of this field

examples/dashboard/dashboard.ipynb
```diff
@@ -0,0 +1,242 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "import hvplot.pandas\n",
+    "import panel as pn\n",
+    "import urllib.request\n",
+    "\n",
+    "pickup_by_zone = pd.read_csv(\"https://github.com/saturncloud/examples/tree/examples-v2.0/examples/dashboard/data/pickup_grouped_by_zone.csv\")\n",
```
jsignell: I don't think this will work as written. I suspect it'd have to be more like: https://raw.githubusercontent.com/saturncloud/examples/examples-v2.0/examples/dashboard/data/pickup_grouped_by_zone.csv 

But I think a public s3 bucket is probably a better solution especially since these files are unlikely to change.

examples/dashboard/dashboard.ipynb
```diff
@@ -0,0 +1,242 @@
...
+    "def trim(target, event):\n",
+    "    target.object = tip_plot(event.new, discrete_slider.value)\n",
+    "\n",
+    "def roll(target, event):\n",
+    "    target.object = tip_plot(date_range_slider.value, event.new)\n",
+    "\n",
+    "discrete_slider.link(tip_timeseries_plot, callbacks={\"value\": roll})\n",
+    "date_range_slider.link(tip_timeseries_plot, callbacks={\"value\": trim})\n",
+    "\n",
+    "with urllib.request.urlopen(\"https://github.com/saturncloud/examples/tree/main/examples/dashboard/data/pickup_map.html\") as f:\n",
```
jsignell: I'd just use relative path for this one.

examples/dashboard/dashboard.ipynb
jsignell: URLs from s3 are like https://s3-us-east-2.amazonaws.com/saturn-public-data/nyc-taxi/data/dashboard/pickup_grouped_by_zone.csv

.gitignore
```diff
@@ -39,24 +39,25 @@ Untitled*.ipynb
...
-*.pkl
-*.pq
-*.rda
-*.Rdata
-*.rds
-*.sqlite
-*.xls
-*.xlsm
-*.xlsx
+## data files
```
jameslamb: in my opinion, checking data into the repo should be the exception, not the rule. This repo is cloned every time Saturn starts up, so growing its size substantially will slow down the start up time for the application (which also means increasing the time-to-recovery during a migration / upgrade).

So instead of allowing these types of files into the repo, I'd prefer to make the default behavior "do not allow data in the repo", and have this PR add specific exclusions for the files you want to add.

```shell
*.csv
!examples/dashboard/data/pickup_grouped_by_time.csv
!examples/dashboard/data/pickup_grouped_by_zone.csv
```

If that suggestion is rejected and we decide to go with "allow any data files in the repo", then I'd like to recommend just removing all these gitignore rules instead of commenting them out. Commented-out code doesn't change the behavior of the software in this repo, but it increases the size of the repo and can lead to false positives from search (like `git grep` or GitHub search).

examples/dashboard/dashboard.ipynb
```diff
@@ -0,0 +1,244 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "import hvplot.pandas\n",
+    "import panel as pn\n",
+    "import urllib.request\n",
```
jsignell, line 31: I think it's easier to just use `open`.

jameslamb (dismissed):

.gitignore
jsignell: conversation is happening in slack about where to store data.

jameslamb: oh ok, I see the conversation now. I'll remove my blocking review

jnolis:
> The changes in this PR require corresponding changes to Saturn itself and to Saturn's integration tests.
> 
> Before I review I want to know...have you and @jsignell talked about that?

Yes, we discussed this!

jnolis:
Okay, as of this morning here is where this PR stands:

## Changes since original PR:

* CPU and GPU examples have been returned
* Any flat files required by the quickstart have been moved to the S3 bucket `saturn-public-data` in the folder `examples/dashboard`
* The `categories` key in the saturn.json has been removed from the readme and weight has been added
* The dashboard notebook was fixed and now correctly runs
* The .gitignore rule to ignore data was readded since that's now going to be in S3

## Open items:

* Thumbnails - Whether the thumbnails should be left in GitHub with an action to move to S3 or put in S3 directly is unresolved. At this point I am okay with putting them in S3, but I'm not sure where in S3 they should go. If we put them in the `saturn-public-data` S3 it feels kind of weird since that's like, part of the website UI which feels different than data. If they go in `saturn-public-files` then I don't have access to this S3 bucket but also do we really want data scientists to have access to and editing prod?

I think besides the thumbnail thing I've covered everyone's comments, but let me know if I've missed anything. Now might be a good time to rereview it.

jnolis:
Also one minor point, my example is failing one test because it has `readme.md` instead of `README.md`. I don't believe any of the data science team is on systems that are case sensitive (we all use windows or mac), so if it's possible to have the test be case insensitive without it causing issues I think that would be good. I'm not sure if that would break something down the line though.

forana:
> Also one minor point, my example is failing one test because it has `readme.md` instead of `README.md`. I don't believe any of the data science team is on systems that are case sensitive (we all use windows or mac), so if it's possible to have the test be case insensitive without it causing issues I think that would be good. I'm not sure if that would break something down the line though.

Can we just change the filename? We'd like to be consistent.

jsignell:
Yeah I am not sure if jupyterhub is case sensitive, but README.md is the standard that I've seen elsewhere.

jnolis:
Sure, I'll update it here to README.md

jsignell:
I spoke with @hhuuggoo yesterday and he is setting up a special bucket on S3 specifically for these quickstart thumbnails. So I think the easiest resolution for now will be to explicitly include `thumbnail_image_url` which will point to an image in the new bucket.

We need to update the schema validator to allow some additional fields to support including project_template options in the saturn.json. To fix this error:
```
'/home/runner/work/examples/examples/examples/dashboard/.saturn/saturn.json' has the following schema issues: {'environment_variables': ['Missing data for required field.'], 'thumbnail': ['Unknown field.'], 'title': ['Unknown field.'], 'weight': ['Unknown field.']}
```

examples/dashboard/.saturn/saturn.json
```diff
@@ -5,8 +5,8 @@
         "disk_space": "4Gi",
         "ssh_enabled": false
     },
+    "environment_variables": {},
     "description": "Deploy a dashboard to Saturn Cloud",
-    "title": "Deploy Dashboards",
```
jsignell: I like the idea of including the title here. We just need to change the validator on the schema.

jnolis:
On the note about the schema validator, it looks like one of the things it's failing is "environmental_variables". Since my code had none I removed that field, but I just put it back with `"environment_variables": {}`. Long term could that be set as an optional field?

examples/dashboard/.saturn/saturn.json
jnolis: Yeah I took it out, thought about it, then just put it back

jsignell:
> Long term could that be set as an optional field?

Yeah we can make it optional _in the future_.

jsignell:
I opened https://github.com/saturncloud/examples/pull/103 to change the schema validation.

hhuuggoo:
Just a note that w.r.t examples, as @jsignell pointed out, Atlas currently by default, will create 1 example project per directory within examples.  We have a few options.

1.  Holding back the next release until we can auto-populate quickstarts in enterprise installations
2.  Disabling example population for users in Atlas, but that means new enterprise users will have a very blank experience
3.  Be ok with creating 3 examples per user instead of 2
4. Since @jnolis is doing this for quickstarts, we can move this to a separate repository, and that won't impact examples at all.

jsignell:
I am not sure how to say this, but make sure the aspect ratio on the screenshot is 2x1 like 

 

from: https://s3-us-east-2.amazonaws.com/saturn-public-files/dashboard-2x1.png

jsignell:
I vote strongly for 1

jnolis:
> I am not sure how to say this, but make sure the aspect ratio on the screenshot is 2x1 like
> 
> 
> 
> from: https://s3-us-east-2.amazonaws.com/saturn-public-files/dashboard-2x1.png

All of the thumbnails I've made for quickstarts are 500px by 250 px and so they should be the right aspect ratio. Once the bucket is online I'll drop them all in there and make a note in the readme about it!

examples/dashboard/.saturn/saturn.json
```diff
@@ -0,0 +1,12 @@
+{
+    "image": "saturncloud/saturn:2021.02.22",
+    "jupyter": {
+        "size": "medium",
+        "disk_space": "4Gi",
```
jsignell: I missed this before, but this needs to be a valid `disk_space` option. Maybe you meant to put `10Gi` or `40Gi`?

jnolis: Why is 4 not valid? I wanted to conserve space since we shouldn't need much I don't think. Is there a minimum required?

jnolis:
Additional changes now make this:

## Changes since original PR:

* CPU and GPU examples have been returned
* Any flat files required by the quickstart have been moved to the S3 bucket `saturn-public-data` in the folder `examples/dashboard`
* The `categories` key in the saturn.json has been removed from the readme and weight has been added
* The dashboard notebook was fixed and now correctly runs
* The .gitignore rule to ignore data was readded since that's now going to be in S3
* Thumbnail moved to S3 and saturn.json changed to have `thumbnail_image_url`

## Open items:

* Size of the quickstart - I set it to be 4gb, figuring that was enough, but @jsignell suggested it be 10 or higher. I'm not sure what the minimum is but I'll change the config to whatever that is.

examples/dashboard/.saturn/saturn.json
jsignell: It's a selector, so there are only certain options. You can see this reflected in the UI.

jsignell:
`"10Gi"` is the minimum. 

I ran this on integration tests https://github.com/saturncloud/release-images/runs/2088941261?check_suite_focus=true with a few tweaks including that. I'll open up a PR now so that you can pull them in.

jsignell:
@jnolis can you also make sure you are good with https://github.com/saturncloud/examples/pull/103

jsignell:
and https://github.com/saturncloud/examples/pull/104 is the PR I mentioned above

README.md
```diff
@@ -35,7 +35,12 @@ The most important file within the `.saturn` folder is the `saturn.json` file wh
 }
 ```
 
-It's possible that other files might existing in the .saturn folder, such as `start_script` which contains the initialization script for the project. However, no file besides `saturn.json` is required.
+It's possible that other files might existing in the .saturn folder, such as `script` which contains the initialization script for the project. However, no file besides `saturn.json` is required.
```
jsignell: `start` not `script` right?

README.md
```diff
@@ -35,7 +35,12 @@ The most important file within the `.saturn` folder is the `saturn.json` file wh
...
+
+Notes about the project structure:
+* The disk_space must be one of the preset choices from the Saturn Cloud UI, it can't be an arbitrary amount of disk space.
+* The startup script must be name `start` without a file extension for Atlas to know it.
+* Options like "environmental_variables" may be required even if they are empty, be aware there is a risk in removing them entirely.
```
jsignell: "environment_variables" :)

jsignell (approved):
:clap:

PR #103, opened 2021-03-11
Update schema to allow project_template fields.

jsignell:
Update schema validator to allow project_template fields.

### DIFF
--- a/.ci/validate-examples.py
+++ b/.ci/validate-examples.py
@@ -73,6 +73,9 @@ class SaturnJsonSchema(Schema):
     dask_cluster = fields.Nested(DaskClusterSchema, required=False)
     required_secrets = fields.List(fields.String(), required=False)
     description = fields.String(required=True)
+    title = field.String(required=False)
+    thumbnail_image_url = field.Url(required=False)
+    weight = fields.Integer(required=False)
 
 
 def image_exists_on_dockerhub(image_name: str, image_tag: str) -> bool:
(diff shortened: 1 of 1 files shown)


.ci/validate-examples.py
```diff
@@ -73,6 +73,9 @@ class SaturnJsonSchema(Schema):
     dask_cluster = fields.Nested(DaskClusterSchema, required=False)
     required_secrets = fields.List(fields.String(), required=False)
     description = fields.String(required=True)
+    title = fields.String(required=False)
+    thumbnail_image_url = fields.Url(required=False)
+    weight = fields.Integer(required=False)
```
jameslamb, line 78: What is a case where we would want to have a quickstart that doesn't have a title, thumbnail, or weight? I have a preference for these being required to force quickstart authors to make those choices.

jsignell: Right. I was trying to accommodate the state we are in right now where we have old style and new style in the same repo. But my intention is for these to be required as soon as quickstarts is "the way"

hhuuggoo: the image isn't really important.  If I'm setting up a bunch of examples for my team, I might not want to think about which image to use.  @jameslamb @jsignell would you be ok with defaulting to the saturn logo in this case?

jsignell: No, we are only talking about quickstarts that we control and that in the future we will automatically add to all new installs. These things should have nice images.

jameslamb: alright I guess I don't have enough context to review this. Is there somewhere that I can read about the plans for the "new style"?

Otherwise maybe someone else should review this

jsignell: All I mean is there is a plan to move from creating example projects for each user, to pre-populating new installs with default project_templates.

jameslamb: ok I can get out of the way

jameslamb (approved):

jnolis (approved):

PR #105, opened 2021-03-11
rapids quickstart

rikturr:
This PR moves over the `examples-gpu/nyc-taxi` content into a dedicated RAPIDS quickstart. The top-level notebooks are super fast intros to RAPIDS with a single GPU then cluster, then there is an `advanced` folder that has the full notebooks with deeper explanations (unchanged from `examples-gpu/nyc-taxi`)


jsignell (dismissed):
This looks good to me. But I'll let @jameslamb give any feedback on content.

jsignell:
Actually let me run this through integration tests real quick... https://github.com/saturncloud/release-images/runs/2124652643?check_suite_focus=true

jsignell:
@rikturr got this error from integration tests:

```python-traceback
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/project/examples/02-rapids-dask-836790fb-e8cc-4b6b-b6f7-78ed80d133b3.py in <module>
     96 
     97 
---> 98 _ = rfc.fit(X, y)
     99 
    100 

/srv/conda/envs/saturn/lib/python3.7/site-packages/cuml/dask/ensemble/randomforestclassifier.py in fit(self, X, y, convert_dtype)
    267         self._fit(model=self.rfs,
    268                   dataset=(X, y),
--> 269                   convert_dtype=convert_dtype)
    270         return self
    271 

/srv/conda/envs/saturn/lib/python3.7/site-packages/cuml/dask/ensemble/base.py in _fit(self, model, dataset, convert_dtype)
    136                 warnings.warn(warn_text)
    137             else:
--> 138                 raise ValueError("Data was not split among all workers. "
    139                                  "Re-run the code or "
    140                                  "use ignore_empty_partitions=True"

ValueError: Data was not split among all workers. Re-run the code or use ignore_empty_partitions=True while creating model

==================== 1 failed, 4 passed in 1210.65 seconds =====================
```

examples/rapids/02-rapids-dask.ipynb
```diff
@@ -0,0 +1,220 @@
+{
```
jameslamb, line 1: <p>I think this needs to be <code>SaturnCluster(n_workers=n_workers)</code>.</p><p> </p><p>Otherwise, if the default for <code>dask-saturn</code> changes to, say, <code>2</code>, this <code>wait_for_workers()</code> code would hang forever. I bet that would take us a while to find :grimacing:</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/saturncloud/examples/pull/105/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

jameslamb (dismissed):
I just left one suggestion based on reviewing this on ReviewNB. I didn't look at anything in `advanced/` since you said that was just copied over.

For the issue in https://github.com/saturncloud/examples/pull/105#issuecomment-800543987, I'm surprised to learn that `cuml` would throw an error if not all workers have a piece of the training data. It seems a bit aggressive to say "if you're not using all workers we raise an exception, but you can disable that with a flag". If passing that `ignore_empty_partitions=True` doesn't work, you could try using `client.rebalance()` (https://distributed.dask.org/en/latest/api.html#distributed.Client.rebalance).

examples/rapids/02-rapids-dask.ipynb
jsignell: This won't use the default, it'll use the setting in saturn.json, but it is brittle to define it in both places.

jameslamb: What are you suggesting? Not using `wait_for_workers()` at all in these notebooks?

rikturr: Sorry I meant to actually remove n_workers and `wait_for_workers()`. We're settling on defining the cluster spec in the `saturn.json` so we shouldn't pass arguments to `SaturnCluster` at all. I think the experience of spinning up a quickstart should include starting a cluster so workers would already be up

jsignell: Well we could pass it in using an env var, or we could not use `wait_for_workers`

rikturr: final decision: 
- in saturn.json, specify cluster config
- in notebooks, do:
```python
n_workers = 3 # same as specified in saturn.json
cluster = SaturnCluster(n_workers=n_workers)
client = Client(cluster)
client.wait_for_workers(n_workers)
```

rikturr:
made changes an running integration test: https://github.com/saturncloud/release-images/runs/2132536640?check_suite_focus=true

jnolis (changes requested):
Overall this looks good! I have two changes and then I'll approve (all with the text, none with the code):

1. It looks like the two workbooks have the same title in the notebooks--can you check that?
2. Can you add a bit more explanatory text? Just like:
  *  "in this notebook we're going to be training a randomforest model on ny taxi data to find..."
  * "This next chunk creates the `RandomForestClassifier` object from the RAPIDS library which does most of the work"
  * etc
  
Nothing too long or explanatory, just enough to give the reader more context on what is happening within each chunk. I wouldn't spend more than 20-30 minutes on it.

Thanks!

rikturr:
Made changes and ran this integration test (test says failed but thats due to a concurrency issue with our testing setup, if you look the individual tests actually all passed) https://github.com/saturncloud/release-images/runs/2133440638?check_suite_focus=true

jsignell (approved):

jnolis (approved):

examples/rapids/02-rapids-dask.ipynb
jameslamb: nice, thanks

jameslamb (approved):
looks good, thanks

PR #108, opened 2021-03-16
Added blank quickstarts

jnolis:
These unassuming blank quickstarts are for getting started as quickly as possible.


examples/blank-gpu/.saturn/saturn.json
```diff
@@ -0,0 +1,13 @@
+{
+    "image": "saturncloud/saturn-gpu:2021.01.13",
```
jsignell: We shouldn't recommend that people use saturn-gpu anymore. It's never going to get anymore updates.

```suggestion
    "image": "saturncloud/saturn-rapids:2021.02.22",
```

jnolis:
After a bit of back and forth between me and @jsignell, we've decided we shouldn't have blank examples because you should set up custom projects for those

PR #109, opened 2021-03-17
Remove all examples-cpu, examples-gpu

jsignell:
Queueing this up for after https://github.com/saturncloud/saturn/pull/2075


rikturr:
This docs page currently points people to the examples-cpu folder: https://www.saturncloud.io/docs/examples/prefect/prefect_cloud/

Ideally we would direct that to point to a prefect quickstart instead but we haven't set that up yet so there will be a gap if we go forward merging this. I suspect there are a couple of other links like that floating around the interwebs just for awareness

jsignell:
Good point @rikturr. Is there a plan to create a prefect quickstart?

rikturr:
> Good point @rikturr. Is there a plan to create a prefect quickstart?

yep its on the list! Hopefully end of this week / early next

rikturr (approved):
I'm good with merging with the understanding that prefect quickstart is coming soon

jsignell:
> I'm good with merging with the understanding that prefect quickstart is coming soon

I talked with @jnolis and we agree to wait on this one, and in the meantime add title and things to the examples-cpu/gpu so that we don't run the risk of losing content in the meantime.

jameslamb:
I want to add a note for whenever this is picked up again...I expect that removing `examples-cpu` and `examples-gpu` would cut the runtime of the integration tests (currently around 1.5 hours) by as much as half.

Something to consider as we weight the benefits of keeping these projects around with duplicated content.

jsignell:
oh thanks for pinging on this @jameslamb. I think all the content has now migrated, so @jnolis maybe the time has come to remove these?

jsignell:
Integration tests running at https://github.com/saturncloud/release-images/runs/2510205635?check_suite_focus=true

jsignell:
Ok I got verbal confirmation from @jnolis. I'm going to merge this now :)

PR #110, opened 2021-03-17
CI updates for template

jsignell:
Make template fields mandatory and check that weight is unique.

Once https://github.com/saturncloud/examples/pull/109 lands, this should pass.


jameslamb (commented):
looks good! I love adding the weights.

I'll review again when this is passing.

jameslamb (approved):

PR #112, opened 2021-03-18
Add @jnolis to default reviewer.

jsignell:
This'll make you get automatically added as a reviewer on every PR that has examples content


jameslamb (commented):
I'm good with this! But I think the only approval that should count towards a merge for this one is @jnolis

jnolis (approved):
This is a really funny PR I like it

PR #113, opened 2021-03-18
prefect quickstart

rikturr:
Notebooks taken as-is from examples-cpu/prefect. Integration tests pass: https://github.com/saturncloud/release-images/runs/2140386110?check_suite_focus=true

@jameslamb don't require review from you, but just pinging you here for awareness that this is the new home for the prefect examples


jnolis (changes requested):
Overall this looks good! I have a few small changes just so we can make it flow into the docs page more easily:

* Can you change the notebook titles to make more sense in the context of a list of articles? Like I think if we took out the "score" part of the titles we can drop them into the docs ui
* Can you make the top cell of each notebook just the title so that it works with the ipynb to md tools? The dashboard folder is a good example of one
* Since we'll want the images to show up when these quickstarts are docs, can you move the images in the notebooks from an image folder in the repo to the saturn-cloud-assets S3 bucket?

Thanks!

jsignell (approved):

jnolis (approved):

PR #116, opened 2021-03-18
Dask quickstart

jnolis:
This quickstart has a basic dask example and links to our docs on how to use dask.


rikturr:
I feel like a "Dask" quickstart might need a bit more than this - is this meant to be "Getting started on Dask" or "Getting started with Dask _on Saturn Cloud_"? I've personally never used gather anywhere except what I've seen for the pytorch work we've done. There's also a lot of good Dask setting started content in the [Dask tutorial](https://github.com/dask/dask-tutorial) so should we just tell people how to launch a cluster in Saturn then point people to that?

jnolis:
> I feel like a "Dask" quickstart might need a bit more than this - is this meant to be "Getting started on Dask" or "Getting started with Dask _on Saturn Cloud_"? I've personally never used gather anywhere except what I've seen for the pytorch work we've done. There's also a lot of good Dask setting started content in the [Dask tutorial](https://github.com/dask/dask-tutorial) so should we just tell people how to launch a cluster in Saturn then point people to that?

Good point - it was meant to be getting started with Dask on Saturn Cloud, so my expectation was people would come in with an understanding of Dask already. It felt a little light to me too. 

I think the existing pytorch quickstart has a good "getting started with dask" notebook (that I think @skirmer made), What if we kept this ultra light notebook in to get started and added that getting started with dask one for people who want more detail?

rikturr:
> Good point - it was meant to be getting started with Dask on Saturn Cloud, so my expectation was people would come in with an understanding of Dask already. It felt a little light to me too.
> 
> I think the existing pytorch quickstart has a good "getting started with dask" notebook (that I think @skirmer made), What if we kept this ultra light notebook in to get started and added that getting started with dask one for people who want more detail?

That would be good, and I know we want to balance with not throwing too much content into this but I think at the minimum it should include (or at the least mention and link out to other material):
- connecting to cluster
- delayed, persist, compute, etc
- dataframes and/or arrays
- ML (I'm on the fence with this one)

I just don't want people to see it and think Dask is only for these abstract low-level list manipulations and does in fact support higher-level work like dataframes and ML

jameslamb (commented):
Seems like there is still some discussion about what this PR will contain (https://github.com/saturncloud/examples/pull/116#issuecomment-802978693). `@` me whenever that conversation is resolved and I'd be happy to review the content.

jnolis:
Okay, after discussing it with @skirmer, we decided that a good first course of action is to leave the super simple example notebook as-is and have it directly link to the doc on [Parallelization in Python](https://www.saturncloud.io/docs/reference/dask_concepts/). In the future we might want to make a more robust example that supercedes the Parallel and Python one. But it's just not clear yet how our customers will end up using these so let's not change too much around now (cc @rikturr )

rikturr (approved):
Sounds good!

examples/dask/start-with-dask.ipynb
```diff
@@ -0,0 +1,128 @@
+{
```
jameslamb, line 1: <p><code>dask_saturn.SaturnCluster()</code> will create a new cluster for you if you don't already have one connected to this resource. So I think that this line can / should be removed --&gt; "If you are having issues with this, you may not have a running Dask cluster on Saturn Cloud".</p><p> </p><p>Could you replace this language with something like the language used for this step in many of the other examples? See, for example, the "Initialize a Dask Cluster" section of https://github.com/saturncloud/examples/blob/47fcb28454a64b892574859c90377a21f49cbc94/examples/examples-cpu/nyc-taxi/xgboost-dask.ipynb.</p><p> </p><p>&lt;hr&gt;</p><p> </p><p>I also think there's a small grammatical error in the third sentence. I'd rewrite it as:</p><p> </p><blockquote>The code creates a function that computes exponents and runs it across a list of inputs in parallel. </blockquote>
<br/>

 _Reply via <a href='https://app.reviewnb.com/saturncloud/examples/pull/116/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

jameslamb, line 1: <p>This isn't exactly accurate. Using <code>delayed</code>  doesn't by itself imply that something will be parallelized. I might write this like this:</p><p> </p><blockquote>You can use the <code>@dask.delayed</code>  decorator to change a regular Python function into a lazily-evaluated function. That means that a function call will return a future, instead of a value.</blockquote>
<br/>

 _Reply via <a href='https://app.reviewnb.com/saturncloud/examples/pull/116/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

jameslamb, line 1: <p>I thought this block was difficult to understand. For example, "take this list of pairs of number and run them..." was confusing to me.</p><p> </p><p>What do you think about this proposed rewrite?</p><p> </p><blockquote>The Dask <code>distributed</code> client comes with several methods for managing collections of such futures. The code below shows how to use a few of these.</blockquote><blockquote><br></blockquote><blockquote>* <code>client.map(f, x)</code> = run function <code>f</code> once per item in <code>x</code> . Returns a list of futures.</blockquote><blockquote>* <code>client.gather(futures)</code> = Given a list of futures, pull them back to the client. Returns a list of delayed results (in this example).</blockquote><blockquote>* <code>client.compute(delayed_results)</code> = Given a list of delayed results, materialize the results on the client. Returns a list of futures (<code>sync=True</code>) or actual function results (<code>sync=False</code>)</blockquote><blockquote>* <code>.result()</code> = Wait for a future to complete. Returns its actual value.</blockquote>
<br/>

 _Reply via <a href='https://app.reviewnb.com/saturncloud/examples/pull/116/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

jameslamb, line 1: <p>Suggest changing <code>dask</code>  to <code>Dask</code>  (uppercase), to match Dasks's branding guidelines, which say that the project / ecosystem should be <code>"Dask"</code>  (uppercase <code>d</code>) : https://marketing.dask.org/en/latest/refer.html.</p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/saturncloud/examples/pull/116/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

jameslamb (dismissed):
I just reviewed this. I only looked at the content that's here and didn't review the scope or whether or not we should have an `examples/dask`.

I left a few suggestions for your consideration, and I think they should be addressed (accepted or rejected) before this example is merged.

examples/dask/start-with-dask.ipynb
jnolis: I didn't know it would create the cluster for you! Neat!

jnolis: Okay looking at the link you sent, I want to avoid going into too much detail about creating the cluster from within the notebook. I think it's confusing that the code creates the cluster once, and it's easier to think about having the cluster creation done though the UI, which I think we lead users towards in our design.

That said, I'll update the notebook to be more clear about exactly what it's doing in this case.

jnolis: What do you think of this? I took what you put and expanded it out a little, although I am not a Dask expert so I might have messed something up

```
The Dask distributed client comes with several methods for managing collections of such futures. The code below shows how to use a few of these.

* client.map(f, x) = run function f once per item in a list-like object x. Returns a list of futures that can then be evaluated.
* client.gather(futures) = Given a list of futures in a Dask cluster, pull them back to the client machine. Returns a list of delayed results (in this example).
* client.compute(delayed_results) = Given a list of delayed results, brings the results on the client machine. Returns a list of futures (sync=True) or actual function results (sync=False)
* .result() = Wait for a future to complete. Returns its actual value.

All together, the code below will take the list of inputs, create lazy evaluations of the exponents on a Dask cluster, gather them onto the client machine, and get the results.
```

jameslamb: I would use `client` instead of `client machine`. It's confusing to say `client machine` because of you use a Dask `LocalCluster` (a really common way to work with Dask), then you're not talking about moving data between *machines* but between worker *processes* and the client *processes*.

Instead of "in a Dask cluster", if you want to add more detail to `client.gather()` it would be more precise to say "Given a list of futures held by Dask workers".

I also realize there was a mistake in my previous comment. `client.compute(sync=True)` returns actual values and `client.compute(sync=False)` returns a list of futures (I had reversed them).

I personally think the phrase "create lazy evaluations of the exponents" is confusing.

jnolis: Okay, here are the additional changes, which I hope should cover it (and I pushed it in the branch too):

```
The Dask distributed client comes with several methods for managing collections of such futures. The code below shows how to use a few of these.

* `client.map(f, x)` - run function `f` once per item in a list-like object `x`. Returns a list of futures that can then be evaluated.
* `client.gather(futures)` - Given a list of futures held by Dask workers, pulls them back to the client. Returns a list of delayed results (in this example).
* `client.compute(delayed_results)` - Given a list of delayed results, brings the results on the client. Returns a list of futures (`sync=False`) or actual function results (`sync=True`)
* `.result()` - Wait for a future to complete. Returns its actual value.

All together, the code below will take the list of inputs, converts them to futures held by Dask workers for apply the exponental function, gathers them onto the client machine, and determines the results.
```

jameslamb: This is not grammatically correct:

> converts them to futures held by Dask workers for apply the exponental function

I suggest changing that sentence to something like this:

> All together, the code below takes a list of inputs, and uses Dask to parallelize the of work executing `lazy_exponent()` on each input.

jameslamb (commented):
I'm going to start my weekend (it's 5:00p here in Central Time). This is close so I'm removing my blocking review. Feel free to merge whenever you feel comfortable with the state of this.

PR #117, opened 2021-03-22
made instances correct for freemium

jnolis:
This PR standardizes all the quickstarts to use either "large" or "g4dnxlarge" for all the instances (including the scheduler). This will make us conform with the freemium requirements, and should make maintaining these easier.

Some of the examples currently will break in freemium since they require instances that are not allowed, so ideally we would roll out a solution fairly quickly.


rikturr (approved):
good catch!

hhuuggoo (approved):

PR #121, opened 2021-04-13
Quick Starts for local-connection cluster use

skirmer:
Probably still needs revision but trying to move it along. Because this is meant explicitly to be run locally, not on Saturn proper, I'm not quite sure what the tests really ought to look like.


hhuuggoo:
Task linked: [DATA-277 Take avidxchange local to dask demo and make it generally available](https://app.clickup.com/t/1250339/DATA-277)

examples/local-access/setup-external-connection.ipynb
```diff
@@ -0,0 +1,202 @@
+{
```
rikturr, line 1: <p>Is it possible to do this with the quickstart project, rather than telling people to create a new one?</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/saturncloud/examples/pull/121/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

rikturr, line 1: <p>I think the settings here should be changed to fit into freemium?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/saturncloud/examples/pull/121/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

rikturr, line 1: <p>Thoughts:</p><ul><li>Is it too much to instruct users to just use our Docker image for this?  </li><li>Should we be hard-coding versions in to the quickstart? What if images change and we forget to update the quickstart? Maybe we can use <code>conda list</code>  to show up-to-date versions</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/saturncloud/examples/pull/121/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

examples/local-access/.saturn/saturn.json
```diff
@@ -0,0 +1,22 @@
...
+    "image": "saturncloud/examples-gpu:2021.02.22",
+    "jupyter": {
+        "size": "g4dnxlarge",
+        "disk_space": "10Gi",
+        "ssh_enabled": false
+    },
+    "dask_cluster": {
+        "n_workers": 3,
+        "nthreads": 1,
+        "scheduler_size": "medium",
```
rikturr, line 11: i think this needs to be large for freemium

examples/local-access/README.md
```diff
@@ -0,0 +1,12 @@
+## Use Cluster from Local Environment Quickstart
+
+In this project, you'll learn how to create and interact with a Dask cluster from your local environment using the Saturn Cloud service. This allows you to skip interacting with the Saturn Cloud UI almost entirely, if you want to.
```
rikturr, line 3: I think it would make sense in the readme to instruct users to do all the things that would need to happen in saturn, then its clear that all the notebooks should be run from their local environment

rikturr (changes requested):
Love this! One thought about naming - saying "local" in the title might not draw the line to that they could run things in other systems like sagemaker, etc. Is there another phrase we could use? or at least clarify in the description

examples/local-access/setup-external-connection.ipynb
skirmer: Could be! Maybe the pytorch quickstart? I didn't want people to get confused about all the notebooks already in the quick start is the problem- if they don't intend to use them, then that seems like it would be distracting at best.

examples/local-access/.saturn/saturn.json
skirmer: I'll check and see what runs, I didn't test it and should've!

examples/local-access/setup-external-connection.ipynb
skirmer: I thought these settings would be allowed but let me test and see- you're probably right!

skirmer: Well, this is an open question- we can try to get people to use our docker image as the kernel for their local jupyter environment, but because it's their local context there may be a lot of unknown unknowns as it were. 

The versions are to try and give them concrete advice about what I know works as far as versioning, since we have libraries pinned to specific versions in our image/s. But you're right that it could cause tech debt for us. I would expect that when we change the pins of versions then I may want to re-test this whole thing and ensure that there are no unexpected effects from that with specific regard to local connections.

jsignell:
closing since this is stale

PR #123, opened 2021-04-22
Example for R Image Users

skirmer:
As discussed, this includes the start script to get the R kernel available for users. No notebooks included, because we assume the user will create their own. 

The image build is all finished thanks to Alex, so this should be the last bit we need!


skirmer:
The linting errors are not relevant to this example, I don't think:
```
1. Every directory two-levels below '/home/runner/work/examples/examples/examples' must have a README.md. None found for '/home/runner/work/examples/examples/examples/examples-r/_img'.
2. No notebooks were found in '/home/runner/work/examples/examples/examples/examples-r' or its subdirectories
```

examples/examples-r/.saturn/saturn.json
```diff
@@ -0,0 +1,15 @@
...
+        "disk_space": "10Gi",
+        "ssh_enabled": false
+    },
+    "dask_cluster": {},
+    "environment_variables": {},
+    "description": "Start a project using R in Jupyter for data science.",
+    "title": "R Project",
+    "thumbnail_image_url": "https://s3-us-east-2.amazonaws.com/saturn-public-assets/example-thumbnails/rlogo.png",
+    "weight": 150,
+    "include_in_every_saturn": false
```
jsignell, line 14: :clap:

jsignell (commented):
I'm wondering if it wouldn't be worth just adding a simple readme to _img like we do in other examples: https://github.com/saturncloud/examples/blob/main/examples/examples-cpu/_img/README.md

Do you think it makes sense to add a simple notebook that has some r code in it and just says look it's R! 

Independent of the validation does that seem like something that would help orient users?

examples/examples-r/README.md
```diff
@@ -0,0 +1,34 @@
+<img src="./_img/saturn.png" width="600">
+
+# Start Using R in Saturn Cloud
+## Welcome to Saturn Cloud!
+
+This project gives you an environment ready for conducting data science in R in Jupyter. Open a new notebook and select the R kernel to begin.
+
+The following libraries are included in the image, and you can install others with `install.packages()`, `devtools`, or `remotes` according to your preference.
```
jameslamb: it's way more than this, right? This is just the list of things we pass to `install.packages()` but you'll get all of their recursive dependencies.

You can get the full list like this:

```shell
IMAGE=$(
     cat .saturn/saturn.json \
     | jq -r ."image"
)
docker run \
    -t ${IMAGE} \
    Rscript -e "options(datatable.print.nrows = Inf); print(data.table::as.data.table(installed.packages())[, .(Package)])"
```

As of this version of the image, that is the following:

```text
          Package
  1:      askpass
  2:    base64enc
  3:           BH
  4:         brew
  5:         brio
  6:       cachem
  7:        callr
  8:        caret
  9:          cli
 10:        clipr
 11:   colorspace
 12:   commonmark
 13:       config
 14:        cpp11
 15:       crayon
 16:  credentials
 17:         curl
 18:   data.table
 19:         desc
 20:     devtools
 21:      diffobj
 22:       digest
 23:        dplyr
 24:     ellipsis
 25:     evaluate
 26:        fansi
 27:       farver
 28:      fastmap
 29:      foreach
 30:           fs
 31:     generics
 32:         gert
 33:      ggplot2
 34:           gh
 35:     gitcreds
 36:         glue
 37:        gower
 38:       gtable
 39:        highr
 40:          hms
 41:    htmltools
 42:         httr
 43:          ini
 44:        ipred
 45:    IRdisplay
 46:     IRkernel
 47:      isoband
 48:    iterators
 49:     jsonlite
 50:        keras
 51:        knitr
 52:     labeling
 53:         lava
 54:    lifecycle
 55:     lightgbm
 56:    lubridate
 57:     magrittr
 58:     markdown
 59:      memoise
 60:         mime
 61: ModelMetrics
 62:      munsell
 63:     numDeriv
 64:      openssl
 65:       pbdZMQ
 66:       pillar
 67:     pkgbuild
 68:    pkgconfig
 69:      pkgload
 70:         plyr
 71:          png
 72:       praise
 73:  prettyunits
 74:         pROC
 75:     processx
 76:      prodlim
 77:           ps
 78:        purrr
 79:           R6
 80:     rappdirs
 81:    rcmdcheck
 82: RColorBrewer
 83:         Rcpp
 84:        readr
 85:      recipes
 86:     rematch2
 87:      remotes
 88:         repr
 89:     reshape2
 90:   reticulate
 91:        rlang
 92:     roxygen2
 93:    rprojroot
 94:   rstudioapi
 95:    rversions
 96:       scales
 97:  sessioninfo
 98:      SQUAREM
 99:      stringi
100:      stringr
101:          sys
102:   tensorflow
103:     testthat
104:       tfruns
105:       tibble
106:        tidyr
107:   tidyselect
108:     timeDate
109:      usethis
110:         utf8
111:         uuid
112:        vctrs
113:  viridisLite
114:        waldo
115:      whisker
116:        withr
117:         xfun
118:      xgboost
119:         xml2
120:        xopen
121:         yaml
122:      zeallot
123:          zip
124:         base
125:         boot
126:        class
127:      cluster
128:    codetools
129:     compiler
130:     datasets
131:      foreign
132:     graphics
133:    grDevices
134:         grid
135:   KernSmooth
136:      lattice
137:         MASS
138:       Matrix
139:      methods
140:         mgcv
141:         nlme
142:         nnet
143:     parallel
144:        rpart
145:      spatial
146:      splines
147:        stats
148:       stats4
149:     survival
150:        tcltk
151:        tools
152:        utils
```

Alternatively, if you're worried about that list being out of date as the image is rebuilt from release to release, then I recommend changing this documentation to something like the following.

> The following packages and all of their `Imports`, `Depends`, and `LinkingTo` dependencies are included in the image.
> To see the full list of installed packages, run the following in an R session in Jupyter.
>
> ```r
>    data.table::as.data.table(installed.packages())[, .(Package, Version)]
>    ```

jameslamb (changes requested):
Looks like this is working! I tested today by adding `saturncloud/saturn-r:2021.04.19` as an external image, then starting a project with start script `Rscript -e "IRkernel::installspec()"`.

I can see the R kernel available by default when I start Jupyter.

image

And some tiny sample code is working as expected.

<details><summary>sample code (click me)</summary>

```r
library(data.table)
library(xgboost)
data(agaricus.train, package = 'xgboost')
train <- agaricus.train

bst <- xgboost(
    data = train$data
    , label = train$label
    , max_depth = 2
    , eta = 1
    , nrounds = 2
    , nthread = 2
    , objective = "binary:logistic"
)
```

</details>

image


I left one suggestion about the list of dependencies in the README.

I'm also +1 on Julia's idea about adding a simple notebook with an example.

For the linting error about `_img`...I thought that the data science team reached a team agreement about storing images for the examples in a public S3 bucket. I don't see that in https://github.com/saturncloud/examples/blob/main/CONTRIBUTING.md but I remember there being a conversation about it in Slack and I see that other examples are getting their images from S3, for example https://github.com/saturncloud/examples/blob/46907612a5734dfcfdea5dd02bb7fe905e7f5c37/examples/prefect/01-prefect.ipynb?short_path=6a6efe4#L17. I think the best way to resolve that error is to upload this image to S3 and source it from there.

examples/examples-r/README.md
skirmer: Good point, I think your second idea is right- say these plus all their imports etc are included and give that code.

skirmer: As for the _img thing I was just following the other examples- I am fine with keeping the image on S3, I have no particular preference! Fewer non-code files on GH is better I suppose!

jsignell: Yeah I like the second option of just mentioning that all the dependencies are also included.

jameslamb: ok cool, cool.

> I was just following the other examples-

ah! Ok I see now I see that `examples-cpu` does still do the `_img/` folder thing (https://github.com/saturncloud/examples/tree/main/examples/examples-cpu/_img), so I guess there isn't an agreed standard for this repo.

Given that, I'm indifferent between S3 and adding the `_img/` folder. Whichever you prefer is cool with me. But I'm opposed to adding an exception for the R example in any of the linting code.

jsignell: Yeah I think the conscensus was that little images like logos are fine to just include in GH. You can totally just copy the _img readme from another example

jnolis: Wait I thought we decided to put them in the s3 `saturn-public-assets` folder. The reason for this is if we ever turn a quickstart into an example doc we don't want the image links to break. I think in this case where it's not a default quickstart that matters less, but I think consistency would be good.

jsignell: Totally up to you all! Sorry for trying to summarize something I half-remembered :grimacing:

jameslamb: This repo already has inconsistencies in how images are handled so I think this discussion should happen away from this PR. This PR wouldn't be introducing any new inconsistency by using an `_img/` folder or by putting the image on S3, and I don't think it needs to be blocked waiting on a bigger decision about how images are handled here.

jnolis (approved):
I have the two minor changes of:

* Move the _img content to S3
* Take @jameslamb's suggestion and use it without the data.table dependency

Since those are both small, I am marking this as approved assuming those changes are made.

examples/examples-r/README.md
jameslamb: sure, good idea

skirmer:
The current linting complaint is that black doesn't know how to read R code. :)

jameslamb:
From https://github.com/saturncloud/examples/pull/123/checks?check_run_id=2420287871

> error: cannot format /home/runner/work/examples/examples/examples/examples-r/check-R-session.ipynb: Cannot parse: 17:49: ggplot(pressure, aes(x=temperature, y=pressure))+

ah!!! I guess it makes sense that `nbqa` with Python-specific stuff wouldn't work with R notebooks. Shoot, I didn't think about that.

I think the config files here should be updated to ignore that R notebook. Let me know if you need help with that, I'd be happy to push something to this branch.

skirmer:
> From https://github.com/saturncloud/examples/pull/123/checks?check_run_id=2420287871
> 
> > error: cannot format /home/runner/work/examples/examples/examples/examples-r/check-R-session.ipynb: Cannot parse: 17:49: ggplot(pressure, aes(x=temperature, y=pressure))+
> 
> ah!!! I guess it makes sense that `nbqa` with Python-specific stuff wouldn't work with R notebooks. Shoot, I didn't think about that.
> 
> I think the config files here should be updated to ignore that R notebook. Let me know if you need help with that, I'd be happy to push something to this branch.

I would appreciate that! I don't want to break anything.

jsignell (approved):
Nice!

jameslamb (approved):
Alright, I got the linting to skip R notebooks!

In case you're curious, after a few minutes of fiddling (unsuccessfully) with the configs for `black` and `flake8`, I found that you can do excludes at the `nbqa` level: https://nbqa.readthedocs.io/en/latest/configuration.html#include-exclude

PR #124, opened 2021-04-30
add CI job to check for broken links

jameslamb:
This PR proposes adding a continuous integration job that will scrape all the external links out of the source code in this repo and check that they're valid.

To be respectful of other sites and to avoid introducing friction into the PR process in this project, this PR proposes that this new check be run once a day, NOT on every pull request.

As of this change, it will also be possible to manually run this job any time you want from https://github.com/saturncloud/examples/actions.

### How to test this

To test manually, pull this branch and run

```shell
make check-links
```

### DIFF
--- a/.ci/check-links.sh
+++ b/.ci/check-links.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+# [description]
+#     look for external links in code, and check if any of them
+#     are broken
+#
+# [usage]
+#
+#     ./pdc/scripts/dev/check-links.sh frontend/src
+#
+
+set -u -o pipefail
+
+SOURCE_TO_CHECK=$1
+
+# get all unique external links, with some exceptions
+#
+# * links to private resources
+# * local resources
+# * links that are templated and require evaluation of code
+#
+URLS=$(
+    grep \
+        -R \
+        -o \
+        --binary-files=without-match \
+        --no-filename \
+        -E '(http|https)://[^"` )>,\\]+' \
+        "${SOURCE_TO_CHECK}" \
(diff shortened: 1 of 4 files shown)


rikturr:
Is there a non-negligible cost with running CI checks daily? I don't really know how the pricing scheme works. I ask just because this repo doesn't really get commits every day so there might be a lot of days where its checking the same links. Would there be a way to run only if the code has changed?

hhuuggoo:
> Is there a non-negligible cost with running CI checks daily? I don't really know how the pricing scheme works. I ask just because this repo doesn't really get commits every day so there might be a lot of days where its checking the same links. Would there be a way to run only if the code has changed?

negligible for us, relative to all the other CI we do.

jameslamb:
>  I ask just because this repo doesn't really get commits every day so there might be a lot of days where its checking the same links. Would there be a way to run only if the code has changed?

Like Hugo said in https://github.com/saturncloud/examples/pull/124#issuecomment-830414176, it's negligible.

But also, the fact that this repo isn't very active is a STRONGER incentive, in my opinion, to run this job daily instead of triggered by code changes. It prevents the situation where, for example, Prefect re-arranges their docs and breaks a link we use in one of the notebooks, and we don't find out about it until a customer tries the notebook.

The goal here is to be notified about broken links before any Saturn users encounter them.

jsignell (approved):

rikturr (approved):

rikturr:
> Prefect re-arranges their docs and breaks a link we use in one of the notebooks, and we don't find out about it until a customer tries the notebook.

Ah yeah good point

PR #126, opened 2021-05-03
 automatically create project, use less aggressive schedule in Prefect Cloud example

jameslamb:
This PR proposes some changes to the Prefect Cloud examples which I think will fix recent issues we've seen in Saturn's integration tests and improve the user experience with these notebooks.

* adds code to ensure that the Prefect Cloud project is created before registering a flow
* changes the schedule for flow runs from "once every 10 minutes" to "once every 24 hours"
* adds documentation on how to manually trigger a single flow run

This should substantially reduce the cost of running the example notebooks and accidentally forgetting to delete the flow or its schedule.


examples/prefect/02-prefect-cloud.ipynb
```diff
@@ -437,7 +467,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "saturn (Python 3)",
+   "display_name": "Python 3",
```
jameslamb:
```suggestion
   "display_name": "saturn (Python 3)",
```

examples/prefect/02-prefect-cloud.ipynb
```diff
@@ -451,7 +481,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.10"
+   "version": "3.7.7"
```
jameslamb:
```suggestion
   "version": "3.7.10"
```

jameslamb:
I just kicked off an integration tests run using this branch.

Won't share the link here since this is a public repo. I'll put the link in Slack, and leave another comment on this PR once the run is complete.

examples/examples-cpu/prefect/prefect-cloud-scheduled-scoring.ipynb
```diff
@@ -103,6 +103,29 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "client = prefect.Client()\n",
+    "client.create_project(project_name=PREFECT_CLOUD_PROJECT_NAME)"
```
bhperry, line 126: will this error if the project already exists?

jameslamb: Nope! I tested it multiple times with the same project name.

And here's where `prefect` explicitly wrote code for handling "already exists" as a success case:

https://github.com/PrefectHQ/prefect/blob/81180ad27de7ae3f0e7418bf7f315093654c1847/src/prefect/client/client.py#L981-L991

examples/prefect/02-prefect-cloud.ipynb
```diff
@@ -108,6 +108,29 @@
     "!prefect auth login -t ${PREFECT_USER_TOKEN}"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<hr>\n",
+    "\n",
+    "## Create a Prefect Cloud Project\n",
```
bhperry, line 117: I did not realize there was another prefect cloud example notebook now. I wonder if that's where all the leftover flows are coming from (because the default integration-tests will just run the notebook, and finish. Doesn't know to disable schedule/delete flow at the end).

This will need an override, like the examples-cpu prefect cloud test has.

examples/examples-cpu/prefect/prefect-cloud-scheduled-scoring.ipynb
bhperry: perfect

examples/prefect/02-prefect-cloud.ipynb
jameslamb: oooo interesting. Ok yeah, this notebook was duplicated over into this place in #113

I can handle the override in the integration tests.

jameslamb:
Ok, the integration tests passed!

The flow was not cleaned up successfully, but that will be a change on the integration tests side and not here.

So I think this is ready to go, as soon as someone from the data science team approves the changes.

bhperry (approved):
LGTM

rikturr (approved):

PR #138, opened 2021-08-06
Draft of the resource template for snowflake ml

skirmer:
Still developing but edits and notes welcome!


skirmer:
Not sure I understand the current linting failure- @jsignell advice?

skirmer:
Once we resolve the linting matter, I think this is ready for review!

jnolis (approved):
I approved this but I have a few minor changes.

Also, I wasn't sure where to put this comment, but I would move `saturncloud-logo.png` to the S3 public bucket so then we don't have to worry about it if we move the notebook (or like if we were to have it copied to the docs)

skirmer:
Ok, got all the notes incorporated!

jsignell:
@forana just pointed out that this PR introduced a 98MB file to this repository. We need to find a different place to put that file. Probably s3.

PR #139, opened 2021-08-06
Remove "www." and fix bad docs link

jsignell:
Mostly fixing one bad docs link, but I also removed "www." just to keep things uniform.


jnolis (approved):
Nice catch on the www.

PR #151, opened 2021-09-08
Use standard snowflake cred env vars

bhperry:
Integration tests failing on this example due to mismatched environment variable names


bhperry:
May not actually be the correct username/password. Does the the SnowflakeM1 example use a different account than the other snowflake example? 

Seeing this error in the integration-tests when I run this branch:
```
DatabaseError: 250001 (08001): Failed to connect to DB: mf80263.us-east-2.aws.snowflakecomputing.com:443. Incorrect username or password was specified
```

skirmer:
Oh, this should not work for any users by default- it is only to be used for the Snowflake Guide that we wrote for them. The names of the snowflake creds are matched to the instructions in that guide, where we tell users to create their own.

Can we omit this from integration tests? It literally should not work, the users must do things on the snowflake side first.

bhperry:
Ah, got it. Yes I will add something to skip this example in the integration-tests.

PR #154, opened 2021-09-27
Tensorflow resource template

skirmer:
Linted, ready for review!


jnolis (approved):
This looks good to me! The TensorFlow png logo wasn't the right aspect ratio, but I just went and fixed it directly in S3 so it should be good to go.

examples/tensorflow/training.ipynb
```diff
@@ -0,0 +1,173 @@
...
+    "    lpath=\"dataset/birds/\",\n",
+    ")"
+   ],
+   "outputs": [],
+   "metadata": {}
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "The first time you run this job, you'll need to download the training and test data in the code chunk above. It'll take a few moments, but it shouldn't be too long. This small sample only includes 61 classes, but the original has 285. "
```
jsignell, line 59: I don't love this sentence: "It'll take a few moments, but it shouldn't be too long." I'd prefer to have an estimate. Something like:  "This will take roughly 2 minutes." 

The time should be fairly consistent I think because it's all happening within AWS and the instance size is the same.

jsignell (commented):
Just a minor comment. Sorry it took me so long to review, I was trying to run it on integration tests and then I got distracted because they were failing for unrelated reasons.

Here are the results from the integration test: https://github.com/saturncloud/release-images/runs/3767484172?check_suite_focus=true

jsignell:
WOOHOO Tests passed!

jsignell (approved):

PR #156, opened 2021-09-28
RAPIDS updates

jnolis:
This PR:

* removes the `nvdashboard` installation from the startup script (https://app.clickup.com/t/1250339/DATA-498 ) since it's now in the base image
* rewords some of the notebooks to make them aligned with current best practices
* fixes an error with JupyterLab and Jupyter Server being out of sync by using a newer version of saturnbase gpu 11.2


jnolis:
I suspect the builds will fail because the new RAPIDS image doesn't exist yet

jsignell (approved):

PR #161, opened 2021-10-06
created api for deployment resource

ms4952:
Based on requirement https://app.clickup.com/t/1250339/DATA-620 .


examples/dashboard/houseprice.py
```diff
@@ -0,0 +1,24 @@
+from fastapi import FastAPI, HTTPException
+import pandas as pd
+from sklearn.linear_model import LinearRegression
+
+app = FastAPI()
+
+
+@app.get("/predict")
+async def predict(BedroomAbvGr: int = 3, YearBuilt: int = 2000):
```
jnolis: I think having defaults as examples in these scenarios is kind of confusing--because if a user forgets to put a parameter in they should get an error, not a prediction with a value they didn't ask for.

examples/dashboard/houseprice.py
```diff
@@ -0,0 +1,24 @@
...
+    df = pd.read_csv(
```
jnolis: I think this line through line 14 should all be outside the predict function call--there is no need to rerun the model fit every time you make a new prediction, and in the case of complex models that could really slow down the API since it would have to constantly be waiting for the model to be retrained.

examples/dashboard/houseprice.py
```diff
@@ -0,0 +1,24 @@
...
+        "https://saturn-public-data.s3.us-east-2.amazonaws.com/examples/dashboard/housePriceData.csv"
+    )
+    lr = LinearRegression()
+    lr.fit(df[["BedroomAbvGr", "YearBuilt"]], df["SalePrice"])
+    a = pd.DataFrame([[BedroomAbvGr, YearBuilt]], columns=["BedroomAbvGr", "YearBuilt"])
+    v = lr.predict(a)
+    if v <= 0:
```
jnolis: Instead of testing if the predicted value is negative, which depends on how the model is fit, I think the test should be: IF NOT ((bedrooms between 0 & 8) AND (Year Built between 1872 & 2100))

jnolis (changes requested):
This looks good! I have a few minor changes for it that should be easy to change.

Also, I think it needs a .saturn folder in it, right? To make the template work?

jnolis (approved):
Looks like you changed all my comments from the previous review. Looks good to me now!

examples/api/.saturn/saturn.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "deployment",
+  "image_uri": "saturncloud/saturn:2021.09.20",
+  "description": "Deploy an API to Saturn Cloud. Read more about deployments at: https://scld.io/docs/deployments.",
```
jsignell, line 4:
```suggestion
  "extra_packages": {
    "pip": "uvicorn fastapi"
  },
  "description": "Deploy an API to Saturn Cloud. Read more about deployments at: https://scld.io/docs/deployments.",
```

PR #163, opened 2021-10-08
Recipes

jsignell:
These are the proposed changes to support https://github.com/saturncloud/saturn/pull/3075 

To Do:
 - [x] change validation in this repo to ensure that recipes are valid


.saturn/templates.json
```diff
@@ -0,0 +1,52 @@
+{
+  "templates": [
+    {
+      "title": "Machine Learning with Snowflake",
+      "thumbnail_image_url": "https://saturn-public-assets.s3.us-east-2.amazonaws.com/example-thumbnails/snowflake.png",
+      "weight": 60,
+      "recipe_url": "https://raw.githubusercontent.com/saturncloud/examples/{REF}/examples/snowflake-ml/.saturn/saturn.json"
```
bhperry: Not entirely sure how I feel about encoding `{REF}` format vars here. Makes sense to have the flexibility, but it also is kind of a bummer that it's not actually a functional URL until you format it.

bhperry: another option that comes to mind would be to just store `"recipe_path": "examples/snowflake-ml/.saturn/saturn.json"`, and have saturn fill in the rest

bhperry: Obviously doesn't make it a functional URL, but it does have the benefit that it could be used in multiple ways. File path, url path, if for some reason someone forked it would still be valid, etc.

jsignell: Yeah I hadn't decided what was best yet. I think I will replace "{REF}" with the ref and make a workflow that updates on release like we do for env_deps in other repos. I like the relative path idea as well though. Maybe that is less fragile.

examples/dashboard/.saturn/saturn.json
```diff
@@ -1,13 +1,23 @@
...
+    "SATURN__JUPYTER_SETUP_DASK_WORKSPACE": "true"
+  },
+  "extra_packages": {
+    "pip": "hvplot"
+  },
+  "working_directory": "/home/jovyan/git-repos/examples/examples/dashboard",
+  "git_repositories": [
+    {
+      "url": "https://github.com/saturncloud/examples",
+      "reference": "release-2021.09.20",
```
jsignell: There's a similar question about what to do with these references. If we want to have the ability to put a button that always gets exactly the same thing then we need to include the real reference in the subdirs. But we should def have an action that updates them.

.ci/validate-examples.py
```diff
@@ -60,38 +55,20 @@ def report(self) -> None:
...
-    jupyter = fields.Nested(JupyterSchema, required=False)
-    deployment = fields.Nested(DeploymentSchema, required=False)
-    dask_cluster = fields.Nested(DaskClusterSchema, required=False)
-    required_secrets = fields.List(fields.String(), required=False)
-    description = fields.String(required=True)
-    title = fields.String(required=True)
-    thumbnail_image_url = fields.Url(required=True)
-    weight = fields.Integer(required=True)
-    include_in_every_saturn = fields.Boolean(required=True)
+    image_uri = recipe["image_uri"]
```
jnolis, line 74: Some other things we might want to manually check in examples specifically:

* All git repos are listed as public
* The instance_types are on the list of ones we allow on the free tier
* Dask clusters have no more than 3 workers

examples/dashboard/.saturn/saturn.json
jnolis: 🤔 This is really tricky. Also if these recipes are going to be public and we want people to look at them, maybe using release branches is bad? Like maybe we should just have `main` be the current best set of recipe examples?

jnolis (approved):
I approved this, but the more I think about the "git references get outdated every release" the more I want to make an passionate plea that our examples should be based on the main branch?

examples/dashboard/.saturn/saturn.json
jsignell: On main these should be "main" and on release branches they should point to a specific release.

bhperry (approved):
LGTM

.ci/validate-examples.py
jsignell: Ok I did 1 and 3. 2 is trickier so I'm going to leave it alone for now.

PR #164, opened 2021-10-08
Removed pipeline example

jnolis:
This PR is to remove the pipeline example--after user testing @jsignell and I noticed this was being a catchall for data scientists, and it's outdated. We're better off not having a catchall example and instead pointing people towards their topic of interest (snowflake, dask, etc)


jsignell:
Let's close this one and open a new PR based off release-2021.09.20 since the Saturn-side changes have merged.

PR #165, opened 2021-10-11
Allow zero notebooks

jnolis:
Currently tests fail if you don't have any notebooks in your example, but now there are cases where we wouldn't have notebooks:

* Template resources for deployments (like APIs)
* R code

### DIFF
--- a/.ci/validate-examples.py
+++ b/.ci/validate-examples.py
@@ -211,13 +211,10 @@ if __name__ == "__main__":
                     )
                     ERRORS.add(msg)
 
-        # every example must have at least one notebook. All cells in that
+        # For all the notebooks (and there may be none), check that all cells in that
         # notebook should be cleared of output and execution counts
         notebook_files = [f for f in all_files if f.endswith(".ipynb")]
-        if len(notebook_files) == 0:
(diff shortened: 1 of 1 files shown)


ms4952 (approved):

.ci/validate-examples.py
```diff
@@ -211,13 +211,10 @@ def _lint_python_cell(file_name: str, code_lines: List[str]) -> List[str]:
...
 
-        # every example must have at least one notebook. All cells in that
+        # For all the notebooks (and there may be none), check that all cells in that
         # notebook should be cleared of output and execution counts
         notebook_files = [f for f in all_files if f.endswith(".ipynb")]
-        if len(notebook_files) == 0:
-            msg = f"No notebooks were found in '{full_dir}' or its subdirectories"
-            ERRORS.add(msg)
-        else:
+        if len(notebook_files) > 0:
```
jsignell: Nitpick, but you don't need this if. You can just go right into the for loop.

jnolis: Nice catch

jsignell (approved):
Horray!

jsignell:
Just a note of do you want this off main or release?

PR #167, opened 2021-10-14
removed seed parameter

ms4952:
cuML regression forest no longer uses parameter seed. Hence have removed that .

### DIFF
--- a/examples/snowflake/advanced/rf-rapids-dask.ipynb
+++ b/examples/snowflake/advanced/rf-rapids-dask.ipynb
@@ -475,7 +475,7 @@
    "source": [
     "from cuml.dask.ensemble import RandomForestClassifier\n",
     "\n",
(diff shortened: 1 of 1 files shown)


hhuuggoo:
just a note that this change will need to be incorporated against release-2021.09.20 (@jnolis not sure if you guys have strong feelings about just re-cutting 09.20, or if you want this PR based against 09.20). but other than that this looks good.

jnolis (approved):

jsignell:
This branch needs to be rebased off release-2021.09.20 and force pushed. Otherwise all the changes that have landed in main since the release branch was cut will accidentally get onto release.

jsignell (changes requested):
Let me know if you need a hand with rebasing @ms4952

jsignell (approved):

PR #175, opened 2021-10-28
Fix ref in templates

jsignell:
I merged #163 before fixing the branch. I still need to work out how this should work in terms of github actions to keep things in sync.


bhperry (approved):
LGTM

I'm gonna plug the idea of only storing relative path again. I think it would simplify things

PR #182, opened 2021-11-05
New action that updates the reference in every recipe.

jsignell:
~~Oooh it almost works. Is it ok with you @forana if you have a bot that pushes straight to release branches like this?~~

This PR adds an action that opens a PR against release whenever a release branch is pushed to. I am not 100% sure if this will trigger when the branch is created, but if not I can change it.

### DIFF
--- a/.ci/update-references.py
+++ b/.ci/update-references.py
@@ -1,0 +1,17 @@
+import json
+import os
+import sys
+
+REF = sys.argv[1]
+
+
+for file in os.listdir("examples"):
+    recipe_path = os.path.join("examples", file, ".saturn/saturn.json")
+
+    with open(recipe_path) as f:
+        recipe = json.load(f)
+        for repo_dict in recipe["git_repositories"]:
+            repo_dict["reference"] = REF
+
+    with open(recipe_path, "w") as f:
+        f.write(json.dumps(recipe, indent=2))
--- a/.github/workflows/update-references.yml
+++ b/.github/workflows/update-references.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Commit new files
         if: ${{ steps.diff.outputs.result }}
         run: |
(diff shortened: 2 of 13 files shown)


jsignell:
I realized that because of our branch protection policies, it needs to be in a PR.

jsignell:
See https://github.com/saturncloud/examples/pull/183 for an example of this action working.

.ci/update-references.py
```diff
@@ -0,0 +1,17 @@
...
+REF = sys.argv[1]
+
+
+for file in os.listdir("examples"):
+    recipe_path = os.path.join("examples", file, ".saturn/saturn.json")
+
+    with open(recipe_path) as f:
+        recipe = json.load(f)
+        for repo_dict in recipe["git_repositories"]:
+            repo_dict["reference"] = REF
```
jnolis, line 14: Will this require us to always include references in the recipes (as in we must include it even if it's `main`)? I think I am okay with that tradeoff I just want to make sure I understand correctly

jnolis (approved):
I had a minor comment but overall I think this is neat!

.github/workflows/update-references.yml
```diff
@@ -0,0 +1,37 @@
+name: update references
+on: create
+
+jobs:
+  update-reference:
+    if: ${{ contains(github.ref, 'refs/heads/release') }}
```
jsignell, line 6: Ok I updated this following this pattern: https://github.community/t/trigger-job-on-branch-created/16878/5 so I think it should happen on branch creation.

.ci/update-references.py
jsignell: no it doesn't imply that. This just makes sure that we _do_ have references on release branches.

bhperry (approved):

PR #188, opened 2021-11-08
Remove reference before opening PR

jsignell:
https://github.com/saturncloud/examples/pull/187 is an example of how the PR opening infra works right now. It'd be nice to automatically push a commit that removes the references again. 

I know this seems like we are going in circles, but it's important that every commit that is on a release branch is also included on main.

### DIFF
--- a/.ci/update-references.py
+++ b/.ci/update-references.py
@@ -2,7 +2,7 @@ import json
 import os
 import sys
 
-REF = sys.argv[1]
+REF = sys.argv[1] if len(sys.argv) == 2 else None
 
 
 for file in os.listdir("examples"):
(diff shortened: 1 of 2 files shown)


bhperry:
What's the intent here?

bhperry:
Oh nvm I think I understand. You're modifying the release->main PR workflow so that it removes the references before going to main. And then anything that uses the recipes on main will get the default reference, which is main.

jsignell:
> You're modifying the release->main PR workflow so that it removes the references before going to main. And then anything that uses the recipes on main will get the default reference, which is main.

Yeah exactly.

bhperry (approved):
LGTM

PR #190, opened 2021-11-10
changed name from job to logger

ms4952:

Current the message to create a job as resource shows job twice (job job) . Hence have changed name  to 'logger'


jsignell (changes requested):
This is great, but can you put it off `release-2021.10.26`

jsignell (commented):
Actually I'll just move it to release-2021.11.10

jsignell (approved):

PR #192, opened 2021-11-11
Add a button to update refs and open a PR

jsignell:
https://github.com/saturncloud/examples/pull/191 is an example of a PR that immediately got stale. I'd like to be able to create a new PR in the UI


jsignell:
eh - not sure this is necessary actually.

PR #195, opened 2021-11-11
Swapped R example for RStudio

jnolis:
This changes the example for R from using Jupyter to using the new RStudio stuff.

At a future point we will want to put it in the `.saturn/templates.json`, but only after the feature is publicly announced.


jnolis:
This is failing linting because of https://github.com/saturncloud/recipes/pull/5

jsignell (approved):

PR #196, opened 2021-11-12
Fix example-job to point to actual dir

jsignell:
We just missed this as part of #190


ms4952 (approved):
oh yess... good catch.

PR #199, opened 2021-11-12
Add a check to make sure that all working_dirs are valid

jsignell:
After #196 I realized there should be a check for this.


jsignell:
You can see what a failure looks like here: https://github.com/saturncloud/examples/runs/4192337989?check_suite_focus=true

ms4952 (approved):

PR #204, opened 2021-11-12
example-prefix

ms4952:
add example- to all name fields in saturn.json


jsignell:
I like it! We should enforce it in CI.

jsignell (approved):

PR #221, opened 2021-12-06
renamed folder for a valid job name

ms4952:
when creating job from a recipe .. I was getting error :'Invalid deployment/job name- since name shouldn't be more than 24 characters , hence changed the name to a shorter one


jnolis:
I noticed several other issues not included in this PR (and this PR renames some folders within the repo which isn't necessary), so I am going to close it and do another

PR #223, opened 2021-12-08
changed the environment variable name

ms4952:
This pr is for autoshut off, where I have changed name of environment variable . The current code is throwing error because of naming of environment variable; since  "Credential names must start and end with a lowercase letter or number, and can only contain lowercase letters, numbers, and dashes."


jnolis:
I think the issue is that the name of the variable _in Saturn Cloud_ has those conditions, but the actual environment variable can be upper case with underscores. Can you try changing the **name** field in Saturn Cloud while leaving the **Environment Variable** the same? if that works I'll close this.

PR #228, opened 2021-12-10
Added the RAPIDS comparison

jnolis:
This is the example to support the collaborative blog post from last August

https://developer.nvidia.com/blog/zero-to-rapids-in-minutes-with-nvidia-gpus-saturn-cloud/


nathanballou (approved):
Looks good

PR #245, opened 2022-01-10
Update templates

jnolis:
This updates the template file in a few ways:

* Removes the RStudio template (we now have a button for that)
* Adds the R: jobs, deployment (dashboard), and deployment (api)
* Rearranges them slightly
* Adds the language to the title of each template.

Some of the template titles are now two lines but I think that's acceptable:

image

**Work for an upcoming PR:** Since now many of the examples tie directly to a docs page (like the two RAPIDS examples are 1:1 what's on the [RAPIDS docs page](https://saturncloud.io/docs/examples/python/rapids/), I am going to update the descriptions so they have links to the docs in them when appropriate. This is important for the jobs/deployments ones since those don't have notebooks you can read.

### DIFF
--- a/.saturn/templates.json
+++ b/.saturn/templates.json
@@ -1,63 +1,81 @@
 {
   "templates": [
     {
-      "title": "Dask",
+      "title": "Dask (Python)",
       "thumbnail_image_url": "https://saturn-public-assets.s3.us-east-2.amazonaws.com/example-thumbnails/dask.png",
       "weight": 100,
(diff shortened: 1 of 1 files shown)


jsignell (commented):
I know this isn't obvious, but this requires a change in saturn because we have some tests in there that make sure that titles are pulled in properly and whatnot. I'll open the corresponding PR.

jsignell (approved):
LGTM

PR #248, opened 2022-01-10
readme files for api and job

ms4952:
We need to convert Python jobs, api pages  into the readmes , to add to  make_md.py for creating doc pages


nathanballou (approved):
LGTM

PR #255, opened 2022-01-11
adding shortlinks

nathanballou:
Added shortlinks to descriptions of api, dashboard, and job examples for both R and Python


jnolis (approved):
Approved, but just to save space can we change all of these to the shorter `Learn more at:` rather than like `Read more about Python API deployments:`. Just saves screen real estate.

Also, can this be rebased to `release-2021.01.06`

PR #257, opened 2022-01-12
Change path to be absolute and called path

jsignell:
Once https://github.com/saturncloud/recipes/pull/9 is merged we can cut a new recipes branch and then this should pass.


jsignell:
Ok this is passing and ready for merge

forana (approved):

PR #259, opened 2022-01-13
Update r-tensorflow.Rmd

jnolis:
I added a numpy dependency a few weeks ago because my code wasn't working, but now it mysteriously works without it, so I'm removing it.


nathanballou (approved):
LGTM

jnolis:
Built off wrong branch

PR #265, opened 2022-01-13
Add versions

jsignell:
This won't pass until https://github.com/saturncloud/recipes/pull/13 is merged, but essentially the point is to include the version in every recipe so that we know what to do with it if it gets passed directly to Saturn.

### DIFF
--- a/.ci/update-references.py
+++ b/.ci/update-references.py
@@ -1,10 +1,9 @@
-from ensurepip import version
 import json
 import os
 import sys
 
 REF = sys.argv[1] if len(sys.argv) == 2 else None
-VERSION = REF.replace("release-","") if REF else None
+VERSION = REF.replace("release-", "") if REF else None
 
 
 for file in os.listdir("examples"):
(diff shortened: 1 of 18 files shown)


.ci/update-references.py
```diff
@@ -15,6 +16,10 @@
                 repo_dict["reference"] = REF
             else:
                 repo_dict.pop("reference", None)
+        if VERSION:
+            recipe["version"] = VERSION
+        else:
+            recipe.pop("version")
```
jsignell: I think this should be part of update, but maybe not? I dunno. More thought needed.

.ci/update-references.py
```diff
@@ -17,4 +17,4 @@
                 repo_dict.pop("reference", None)
 
     with open(recipe_path, "w") as f:
-        f.write(json.dumps(recipe, indent=2))
+        f.write(json.dumps(recipe, indent=2) + "\n")
```
jsignell, line 20: This was just bugging me.

forana (approved):

PR #266, opened 2022-01-14
remove `concurrency_policy` and `backoff_limit` from non-scheduled job recipe

forana:
We have a known issue in the recipe schema that says that these two fields are allowed for non-scheduled jobs - that is not correct. For a non-scheduled job, these fields must not be present.


jnolis (approved):
This makes sense!

PR #269, opened 2022-01-14
update examples' images for 2022.01.06 release

forana:
This PR updates all examples' images to use the newly-built 2022.01.06 images.


jsignell (approved):
Just need to update the base branch I think

PR #282, opened 2022-01-25
Fix dask example

jsignell:
There were some issues with this example that came up when trying to run integration tests.

Here is the latest integration test run for this example:

https://github.com/saturncloud/release-images/runs/4940887921?check_suite_focus=true

### DIFF
--- a/examples/dask/.saturn/saturn.json
+++ b/examples/dask/.saturn/saturn.json
@@ -2,8 +2,8 @@
   "name": "example-dask",
   "image_uri": "public.ecr.aws/saturncloud/saturn:2022.01.06",
   "description": "Samples of code and explanations for common Dask user tasks",
-  "environment_variables": {
-    "SATURN__JUPYTER_SETUP_DASK_WORKSPACE": "true"
+  "extra_packages": {
+    "pip": "lightgbm"
   },
   "working_directory": "/home/jovyan/examples/examples/dask",
   "git_repositories": [
--- a/examples/dask/.saturn/saturn.json
+++ b/examples/dask/.saturn/saturn.json
@@ -9,7 +9,7 @@
   "git_repositories": [
     {
       "url": "https://github.com/saturncloud/examples",
-      "reference": "release-2022.01.06"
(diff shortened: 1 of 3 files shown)


examples/dask/.saturn/saturn.json
```diff
@@ -2,14 +2,14 @@
...
-    "SATURN__JUPYTER_SETUP_DASK_WORKSPACE": "true"
+  "extra_packages": {
+    "pip": "lightgbm"
   },
   "working_directory": "/home/jovyan/examples/examples/dask",
   "git_repositories": [
     {
       "url": "https://github.com/saturncloud/examples",
-      "reference": "release-2022.01.06"
+      "reference": "fix-dask"
```
jsignell: I'll change this back before merging, but this is so that integration-tests pick up the right versions of the notebooks.

jsignell: Ok test passed! Switching this back

examples/dask/starting-dask.ipynb
```diff
@@ -111,7 +111,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "client.cluster.close()"
+    "# client.cluster.close()"
```
bhperry: Was this intended to be left commented out?

jsignell: Yeah this is intentional. I don't think we want to close the cluster when people have more notebooks to look at in the example

jnolis: I think that makes sense, but let's convert it from a comment in a code chunk to in the text have it
```
Once you are done using Dask, you can shut down the cluster using the command: `client.cluster.close()`.
```

bhperry: should it be removed entirely then? Otherwise maybe there should be a comment explaining it

bhperry: @jnolis beat me to it. I like that idea ^^

bhperry (approved):
LGTM

PR #284, opened 2022-01-25
Rolling averages Dask example

ms4952:
Adding rolling average dask article to dask special topics


examples/dask/special-topics-rolling-averages.ipynb
```diff
@@ -0,0 +1,197 @@
...
+     "iopub.execute_input": "2022-01-25T17:33:22.602911Z",
+     "iopub.status.busy": "2022-01-25T17:33:22.602324Z",
+     "iopub.status.idle": "2022-01-25T17:33:25.666206Z",
+     "shell.execute_reply": "2022-01-25T17:33:25.665688Z",
+     "shell.execute_reply.started": "2022-01-25T17:33:22.602827Z"
+    },
+    "tags": []
+   },
+   "source": [
+    "\"I need to calculate a rolling average of a numerical column, in time series data. In pandas, I can do this with rolling(x).mean() with sorted values, but what do I do in Dask, with distributed data?\"\n",
```
jnolis:
```suggestion
    "> I need to calculate a rolling average of a numerical column, in time series data. In pandas, I can do this with rolling(x).mean() with sorted values, but what do I do in Dask, with distributed data?\n",
```

examples/dask/special-topics-rolling-averages.ipynb
```diff
@@ -0,0 +1,197 @@
...
+   "source": [
+    "import dask\n",
+    "\n",
+    "timeseries = dask.datasets.timeseries()\n",
+    "timeseries.rolling(\"1D\").mean().compute()\n",
+    "\n",
+    "timeseries = dask.datasets.timeseries()\n",
+    "timeseries.rolling(\"1D\").mean().compute()"
+   ]
+  },
```
jnolis: I don't think there is a need to show a single node example (we don't ever do that on any other example)

```suggestion
```

examples/dask/special-topics-rolling-averages.ipynb
```diff
@@ -0,0 +1,197 @@
...
+    "cluster = SaturnCluster(\n",
+    "    scheduler_size=\"medium\",\n",
+    "    worker_size=\"xlarge\",\n",
+    "    n_workers=3,\n",
+    "    nthreads=4,\n",
+    ")\n",
+    "client = Client(cluster)\n",
+    "client"
+   ]
+  },
```
jnolis: This should _exactly_ match how we do it on the other examples, including the text before and after:

image

examples/dask/special-topics-rolling-averages.ipynb
```diff
@@ -0,0 +1,197 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "280f72b9-9602-460d-9aa0-9caf3999773b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import s3fs\n",
```
jnolis: I think this and the next cell should be merged together. It should have text like:

"First we download the data from S3 and read it into a dask dataframe"

examples/dask/special-topics-rolling-averages.ipynb
```diff
@@ -0,0 +1,197 @@
...
+    "\n",
+    "s3 = s3fs.S3FileSystem(anon=True)\n",
```
jnolis: Why are we using s3fs here? I think we should be doing it the same way we do it in the dask dataframe example right? Where we just directly use read_csv into the dask dataframe? If there is a reason we have to use s3fs we should say why. My hunch is this is just an older example though.

examples/dask/special-topics-rolling-averages.ipynb
```diff
@@ -0,0 +1,197 @@
...
+   "outputs": [],
+   "source": [
+    "taxi = taxi.set_index(\"tpep_pickup_datetime\")\n",
+    "from dask.distributed import wait\n",
+    "\n",
+    "taxi = taxi[\"2019-01-01\":\"2020-01-01\"]\n",
+    "taxi = taxi.persist()\n",
+    "_ = wait(taxi)"
+   ]
+  },
```
jnolis: There should be text describing what this chunk is doing.

also the import statement should be at the top.

examples/dask/special-topics-rolling-averages.ipynb
```diff
@@ -0,0 +1,197 @@
...
+   "outputs": [],
+   "source": [
+    "taxi_new = taxi.join(rolling_fares_df, how=\"outer\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "9fde3376-1f84-4627-852c-f83fabdfc2bf",
+   "metadata": {},
```
jnolis: What does this part show? We should describe it and talk about the conclusion.

jnolis (changes requested):

examples/dask/special-topics-rolling-average.ipynb
```diff
@@ -0,0 +1,179 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8a0e7f9e-8c67-4cd0-adf4-1d082addb1c5",
+   "metadata": {},
+   "source": [
+    "After running the above command, it's recommend that you check on the Saturn Cloud resource page that the Dask cluster as fully online before continuing. Alternatively, you can use the command `client.wait_for_workers(3)` to halt the notebook execution until all three of the workers are ready.\n",
+    "\n",
+    "Now load your data into a Dask Dataframe. Here we are loading data for csv file located at s3 storage. Using read_csv from Dask takes the same form as using that function from pandas."
```
jnolis:
```suggestion
    "Now load your data into a Dask DataFrame. Here we are loading data for csv file located in a public location hosted by Saturn Cloud. Using `read_csv` from Dask takes the same form as using that function from pandas."
```

examples/dask/special-topics-rolling-average.ipynb
```diff
@@ -0,0 +1,179 @@
...
+    "    storage_options={\"anon\": True},\n",
+    ").sample(frac=0.1, replace=False)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a2886561-f4be-481d-95c1-82997a609ae9",
+   "metadata": {},
+   "source": [
+    "For dataframe we are calling function `set_index` to set `tpep_pickup_datetime` columns as the index. This will sort data by index within and across partitions. At this point, data and labels are lazy collections. They won’t be read into the workers' memory until some other computation asks for them. Hence, before we pass this data to other tasks we will call `presist()`. This will ensure that loading of data is run only once."
```
jnolis:
```suggestion
    "For the DataFrame we are calling function `set_index` to set `tpep_pickup_datetime` columns as the index. This will sort data by index within and across partitions. At this point, data and labels are lazy collections. They won’t be read into the workers' memory until some other computation asks for them. Hence, before we pass this data to other tasks we will call `presist()`. This will ensure that loading of data is run only once."
```

examples/dask/special-topics-rolling-average.ipynb
```diff
@@ -0,0 +1,179 @@
...
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "from dask.distributed import wait\n",
+    "\n",
+    "taxi = taxi.set_index(\"tpep_pickup_datetime\")\n",
+    "taxi = taxi.persist()\n",
+    "_ = wait(taxi)"
```
jnolis: Can you add an explanation of why we use this (do we need it?)

examples/dask/special-topics-rolling-average.ipynb
```diff
@@ -0,0 +1,179 @@
...
+   "execution_count": null,
+   "id": "6d6f4e2a-1eda-429e-be99-6e268cac24b1",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "rolling_fares = taxi.fare_amount.rolling(\"30D\").mean()\n",
+    "rolling_fares_df = rolling_fares.to_frame(name=\"fare_amount_rolled\")\n",
+    "type(rolling_fares_df)"
```
jnolis: We dont need this `type...` line, right?

examples/dask/special-topics-rolling-average.ipynb
```diff
@@ -0,0 +1,179 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6d65607a-5629-434a-a876-b5fda59334c4",
+   "metadata": {},
+   "source": [
+    "We then join the original dataframe `taxi` with dataframe `rolling_fares_df`."
```
jnolis:
```suggestion
    "We then join the original DataFrame `taxi` with DataFrame `rolling_fares_df`."
```

examples/dask/special-topics-rolling-average.ipynb
```diff
@@ -0,0 +1,179 @@
...
+    "taxi_new = taxi.join(rolling_fares_df, how=\"outer\")\n",
+    "taxi_new"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "fa9e4e3c-3bba-49ce-9f90-17f775701ae7",
+   "metadata": {},
+   "source": [
+    "On running the cell above, you will see that now we have a new column `fare_amount_rolled` added to dataframe. "
```
jnolis: Can you, either via a screenshot hosted on S3 or in a markdown table, show what like 5 rows of the output look like? It would be good to just see that it works without having to actually run the code.

jnolis (changes requested):
A few small changes! I would like to see it one more time before merging too

jnolis (approved):

PR #292, opened 2022-01-27
Migrated new Dask examples from 2022.01.06 to 2021.11.10

jnolis:
Currently hosted only has 2021.11.10 on it. We have a set of new dask examples that we want to share with our webinar. This will have them in the right version of the schema to run.


examples/dask/.saturn/saturn.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "dask",
-  "image_uri": "saturncloud/saturn:2021.11.10",
-  "description": "Use distributed computing with Dask",
-  "environment_variables": {
-    "SATURN__JUPYTER_SETUP_DASK_WORKSPACE": "true"
+  "name": "example-dask",
+  "image_uri": "public.ecr.aws/saturncloud/saturn:2022.01.06",
```
jsignell: We should probably use the old image: 
```suggestion
  "image_uri": "saturncloud/saturn:2021.11.10",
```

jsignell (commented):
This looks good to me, but let's just use the old image. I can manually replace the quickstart on hosted if you need.

jsignell (approved):

PR #294, opened 2022-01-27
Clean up dask imports

jsignell:
Removing duplicated imports and using `dd` directly instead of importing from `dask.dataframe`

Integration test: https://github.com/saturncloud/release-images/runs/4968462324?check_suite_focus=true

Integration tests seem to not be working on this branch for unrelated reasons, but I tested this on hosted and it seems fine.


jnolis (approved):
Makes sense to me!

PR #301, opened 2022-01-28
correct example working_directory

forana:
Turns out, this is a very subtle way to make our `integration-tests` break. We should probably add a CI check for this at some point.


jsignell (approved):
We should def have a check for this yes.

PR #320, opened 2022-02-21
R future

nathanballou:
R futures examples. Contains base future example, furrr, and targets examples.

The dataset might need to be reduced to improve runtime.


cspell.json
```diff
@@ -0,0 +1,9 @@
+{
+    "version": "0.2",
+    "ignorePaths": [],
+    "dictionaryDefinitions": [],
+    "dictionaries": [],
+    "words": [],
+    "ignoreWords": [],
+    "import": []
+}
```
jsignell: I can't see where this is being used?

examples/r-future/functions.R
```diff
@@ -0,0 +1,87 @@
...
+download_data <- function() {
+    if (!file.exists("births_data.rds")) {
+        download.file(
+            "https://saturn-public-data.s3.us-east-2.amazonaws.com/birth-data/births_2005.rds",
+            "births_data.rds"
+        )
+    }
+    births_raw_data <- readRDS("births_data.rds")
+
+    return(births_raw_data)
```
jnolis: R style guide says to not use the unnecessary return (this applies to I think all of your functions here)

https://style.tidyverse.org/functions.html?q=return#return

```suggestion
    readRDS("births_data.rds")
```

examples/r-future/functions.R
```diff
@@ -0,0 +1,87 @@
...
+}
+
+filter_data <- function(births_raw_data) {
+    births_data <- births_raw_data %>%
```
jnolis, line 19: Do you know where the `%>%` is indirectly being loaded from? It might be better to explicitly load the `library(magrittr)` package which is where it comes from.

examples/r-future/functions.R
jnolis: Also, in general it's better to use `read_rds` and the equivalent functions from the `readr` package when you can. can we add a `library(readr)` and switch the function?

examples/r-future/r-furrr.Rmd
```diff
@@ -0,0 +1,210 @@
...
+#### Run the Model with furrr
+
+Instead, let's try it with a parallel back-end. In this case, we ask for eight workers for eight processes.
+
+The only change to the purrr code is to add the futures `plan` and exchanging `pmap_dfr` with `future_pmap_dfr`. 
+
+```{r furrr evaluation}
+#####################
+## Run this instead##
+#####################
```
jnolis:
```suggestion
######################
## Run this instead ##
######################
```

examples/r-future/r-furrr.Rmd
```diff
@@ -0,0 +1,210 @@
...
+plan(multisession(workers = 8))
+
+print("With furrr:")
```
jnolis:
```suggestion
message("With furrr:")
```

In general it's better to use `message` than `print` for like, logging progress. Message is better because you can use commands like `suppressMessages()` and it's handled differently sometimes when other functions are calling subfunctions.

(there are a few other places you use print too that you might want to swtich)

examples/r-future/r-furrr.Rmd
```diff
@@ -0,0 +1,210 @@
...
+results_furrr <- future_pmap_dfr(params, create_model_table, .options = furrr_options(seed = NULL))
+toc()
+
+best_result_furrr <- results_furrr %>%
+    top_n(-1, mean_absolute_error) %>%
+    head(1)
+
+print(best_result_furrr)
+
+plan(sequential)
```
jnolis:
```suggestion
```

I actually think it's a best practice to not change it back to sequential after you're done. In a sense if you want sequential you just would use purrr instead of furrr

examples/r-future/r-furrr.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+Instead, let's try it with a parallel back-end. In this case, we ask for eight workers for eight processes.
+
+The only change to the purrr code is to add the futures `plan` and exchanging `pmap_dfr` with `future_pmap_dfr`. 
+
+```{r furrr evaluation}
+#####################
+## Run this instead##
+#####################
+plan(multisession(workers = 8))
```
jnolis: I would move this line much higher up and make a section that's like "here we set up future/furrr to use multiple workers by creating 8 R sessions. You can also choose to not set the number of workers and have R guess based on the number of cores"

This line really is like a preamble set up thing.

jnolis: (and I noticed you did it that way in the other file! Which is great!)

examples/r-future/r-future.Rmd
```diff
@@ -0,0 +1,147 @@
...
+1. The execution environment
+2. The expression to the executed
+3. The status of the execution
+
+## The Execution Environment
+
+To start using futures, first specify the execution environment. The execution environment determines where to execute the calculation defined by the future. It is defined using the `plan` function. Consider this example:
+
+```{r plan example, eval = FALSE}
+plan(multisession)
```
jnolis, line 22: Here you didn't set the number of workers but in the other file you did. Can you keep them consistent? I think there is a good argument for either.

examples/r-future/r-future.Rmd
```diff
@@ -0,0 +1,147 @@
...
+* **multicore**: A parallel, non-blocking execution environment -- Instead of running on background R sessions, like multisession, this `plan` runs on forked R processes. This approach has some speed advantages, particularly a reduction in memory overhead, but it is not stable on Windows or RStudio. 
+* **cluster**: A parallel, non-blocking execution environment -- It allows futures evaluations to run on external R sessions. It can be used on both local and remote machines. 
+
+In addition to these built-in plans, futures can also be used with [batchtools](https://github.com/mllg/batchtools), [callr](https://cran.r-project.org/web/packages/future.callr/index.html), or custom backends.
+
+## The Expression to Be Executed
+Futures can be created either implicitly or explicitly. Both styles work to the same end but with different syntax: implicit futures are more similar to regular R code, whereas explicit futures can be clearer and easier to read. 
+
+To define an implicit future, simply replace `<-` with `%<-%`, like in the example below. The value of an implicitly defined future is returned automatically.
```
jnolis, line 39: It might be worth noting that this conflicts with the `%<-%` from the somewhat popular zeallot R package which is used by other packages like keras for returning tuples from functions, so watch out

examples/r-future/r-future.Rmd
```diff
@@ -0,0 +1,147 @@
...
+
+```{r implicit future example, eval = FALSE}
```
jnolis, line 43: Are these eval=FALSE chunks going to mess up our website examples code section? Do we have a check in the regex that removes them?

examples/r-future/r-future.Rmd
```diff
@@ -0,0 +1,147 @@
...
+z <- future(add(x, y), seed = TRUE)
+
+resolved(z)
+print(value(z))
+```
+
+Before ending your session, revert the plan to sequential computations:
+```{r sequential plan reset}
+plan(sequential)
+```
```
jnolis: I would remove this one end sequential in this file too (see my previous note)

jnolis (approved):
I think this is great! I have a few small comments but otherwise it's good to go.

Also, I didn't see a PR for the website to add the example in. I assume that'll happen once this is merged?

examples/r-future/.saturn/saturn.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "example-r-future",
-  "image_uri": "saturncloud/saturn-rstudio-parallel:2022.01.06",
+  "image_uri": "saturncloud/saturn-rstudio-parallel:2022.03.01",
```
jsignell: This PR is off release-2022.01.06 so I don't think that image is available.

forana: Examples are tied to schema versions and not Saturn release versions, right? I don't think there's a restriction, here. That image exists, though it's on ECR: `public.ecr.aws/saturncloud/saturn-rstudio-parallel:2022.03.01`

jsignell: Yeah that's true I guess. Just maybe a slightly odd thing to do, but you are right this is fine.

jsignell (approved):

PR #322, opened 2022-02-21
Add checker for instance types

jsignell:
This PR adds a checker to ensure that all the instance_types mentioned in the recipes in this repo are valid


forana (approved):
LGTM

jnolis (approved):
This looks good to me. A v2 of this would maybe be to check if the type is included in the free version specifically, but I can see some hiccups with that which would make it not worth the effort

PR #323, opened 2022-02-25
Fixed a size issue with the tensorflow

jnolis:
@ms4952 - can you check this still runs?


jnolis:
This will be made in a separate PR

PR #325, opened 2022-02-28
examples supporting practical dask presentation

nathanballou:
Adds examples for the using Dask in practice presentation. Two separate folders for the two separate recipe specifications. Uses shared folders in it's current iteration, so we may need to make modifications.


jnolis (approved):
I'm not going to do a deep review because this really is just to support your presentation so you have a broad flexibility here. That said it is failing the tests so that should be fixed.

jsignell:
Is this still in progress or can we close?

PR #328, opened 2022-03-02
Update images to 2022.03.01

forana:
Updating the images to the newly-built `2022.03.01` tags.


jsignell:
I have thought about this a bit and I think we should run integration tests against this branch to get a sense whether or not this is safe to merge.

jsignell (approved):

PR #338, opened 2022-03-18
update SSH instructions for load files

nathanballou:
Changes the ssh instructions from Credentials to Settings in the examples


jnolis (changes requested):
Can we remove these instructions entirely and just link to the doc page that describes setting up the SSH key? This is both redundant and hard to maintain

jnolis:
we may want to do that in the R (and eventually Julia) examples as well

jnolis (approved):

PR #343, opened 2022-03-24
Update contributing doc and buttons on readme

jsignell:
I was talking through different scenarios with Alex and thought it might be helpful to write down what I see as the plan for this repo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,16 +78,18 @@ After new images are created:
 
 1. Create a pull request into `main` which updates all of the `"image_uri"` entries in `saturn.json` files to the latest images released from https://github.com/saturncloud/images.
 2. Run integration tests using that branch.
-3. If the images are not yet available on one of the staging instances, ask someone on the devops team to update the images there.
-4. Manually test any examples that you are worried about.
-5. Once you are satisfied, ask someone on the devops team to update the images on hosted.
-6. Merge the pull request.
+3. Merge the pull request.
+4. Ask someone on the devops team to update the examples one of the staging instances.
+5. Manually test any examples that you are worried about.
+6. Once you are satisfied, ask someone on the devops team to update the images on hosted.
+
+NOTE: In the future we expect to add a separate mechanism for updating the images separate from the update examples mechanism. Once this is established you will be able to ask someone to update the images on a staging instance then you will be able to test the example using a recipe _before_ merging the PR.
 
 ## Use new version of the recipe schema (rare)
 
 Anytime _after_ a new release makes it to [hosted](https://app.community.saturnenterprise.io), you can start using the latest recipe schema that is supported by that release. Note that the only reason to do this is if you want to take advantage of some new feature in the latest recipe schema.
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -66,27 +66,46 @@ If this raises any issues, try automatically fixing them
...
-3. On `main` update RECIPE_SCHEMA_VERSION to the version on [hosted](https://app.community.saturnenterprise.io).
-4. Update all the `schema_version` fields in all the recipes and make any changes.
-5. Update the button on README.md to point to the new RECIPE_SCHEMA_VERSION
-6. Create a pull request with these changes with `main` as the base.
+1. Create a pull request into `main` which updates all of the `"image_uri"` entries in `saturn.json` files to the latest images released from https://github.com/saturncloud/images.
+2. Run integration tests using that branch.
+3. If the images are not yet available on one of the staging instances, ask someone on the devops team to update the images there.
+4. Manually test any examples that you are worried about.
+5. Once you are satisfied, ask someone on the devops team to update the images on hosted.
+6. Merge the pull request.
```
forana: This part can't work without further modifications, like we discussed yesterday - the examples _are_ how the images get updated.

```suggestion
3. Merge the pull request.
4. If the images are not yet available on one of the staging instances, ask someone on the devops team to update the examples there.
5. Once you are satisfied, ask someone on the devops team to update the examples on hosted.
```

This modification would work with what we have now. Even if we implement the ability for Saturn to come built-in with more images (which I do think we should do), this would require a new set of Saturn builds - not a simple update operation.

CONTRIBUTING.md
```diff
@@ -66,27 +66,46 @@ If this raises any issues, try automatically fixing them
...
 
-This repo has branch protection to prevent deletion or changes without a pull request on all `release-*` branches.
+## Use new version of the recipe schema (rare)
+
+Anytime _after_ a new release makes it to [hosted](https://app.community.saturnenterprise.io), you can start using the latest recipe schema that is supported by that release. Note that the only reason to do this is if you want to take advantage of some new feature in the latest recipe schema.
+
+First create a new branch from `main`. It's name must start with `release-`.
```
forana:
```suggestion
First create a new branch from `main`. Its name must start with `release-`.
```

CONTRIBUTING.md
jsignell: yes sorry I should have clarified what I am trying to write in this document. I am trying to write about an ideal future state where we _do have_ separate "make image-examples" "make examples". There are two reasons for this. One is that there are examples that exist in this repo that aren't part of the default resource templates that we call "examples" in saturn. Those examples might depend on images that are different than the default resource templates. So I think we need some way of saying "these are all the images that any example will need" on hosted. And we need a way to update those images.

jnolis:
Just checking if this is ready to merge?

jnolis (approved):

jsignell:
Umm I should check with @forana again to make sure he's happy with this idea. Probably need to call out the parts that aren't quite there yet too.

forana (approved):
Yep, fine with this for now. We can update as needed.

PR #344, opened 2022-03-28
Bodo examples

nathanballou:
Adds bodo examples. These consist of both scripts and notebooks using the bodo.ai library.


jnolis:
Let's close this and use the other one

PR #345, opened 2022-03-29
Bodo examples (Simplified)

jnolis:
This adjusts the original bodo examples to reduce the number drastically, have the titles and descriptions more align with our other examples.

The example still loads a ton of data 8 minutes worth, and so ideally as a future PR we would change to smaller datasets.

PR #348, opened 2022-03-30
Add disk_space to example workspaces missing it, added validation check

forana:
When #345 merged, this added the Bodo example to the templates lists, which broke our internal CI tests on another repo because, while the 2022.01.06 resource schema technically does not require `disk_space` on jupyter or rstudio servers, our applications currently require it to be set. This PR adds a validation check for that to this repo, and adds disk_space to the two examples that were missing it.


jsignell (approved):
Thanks @forana for fixing!

PR #351, opened 2022-04-05
Added two more examples, one EDA and one TPCH, Updated the current one for better display

alifarhid:
1. The current Bodo-model-training for nye-taxi updated for pretty output display. Deleted the old one and renamed bodo-model-training-nyc-taxi.ipynb

2. Two more examples added. 
2.1. A data processing example showing exploratory data analysis on Chicago crimes.
2.2. TPCH benchmarks


jnolis:
The linter isn't running on this example because the last commit was made by someone outside of Saturn Cloud, so I made #356 which also renamed one of the files and removed the output cells. It now passes testing so I'll merge that one instead.

PR #360, opened 2022-04-25
Update bodo

nathanballou:
This updates the bodo examples to use parquet files instead of csv files. This allows us to set `anon = True` when accessing the files so that the examples are not dependent on AWS credentials being set. 

Also added cluster stopping to two of the examples to be sure that running the examples sequentially doesn't create too many processes.

File locations were fixed on the back end.

### DIFF
--- a/examples/bodo/bodo-tpch.ipynb
+++ b/examples/bodo/bodo-tpch.ipynb
@@ -15,9 +15,6 @@
     "\n",
     "Dataset size is 2GB.\n",
     "\n",
-    "There's a larger dataset available on \"s3://bodo-example-data/tpch/s4/\" which is 4GB. \n",
-    "\n",
-    "\n",
     "**The Bodo parallel cluster in this example runs within the same Saturn Cloud resource as the notebook.** Thus, to increase the performance of the Bodo cluster you only need to increase the instance size of the Jupyter Server resource it's running on.\n",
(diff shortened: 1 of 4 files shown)


examples/bodo/bodo-tpch.ipynb
```diff
@@ -51,31 +48,19 @@
...
-     "iopub.execute_input": "2022-04-04T23:48:34.551756Z",
-     "iopub.status.busy": "2022-04-04T23:48:34.551374Z",
-     "iopub.status.idle": "2022-04-04T23:48:44.880272Z",
-     "shell.execute_reply": "2022-04-04T23:48:44.879646Z",
-     "shell.execute_reply.started": "2022-04-04T23:48:34.551724Z"
+     "iopub.execute_input": "2022-04-25T18:03:43.616460Z",
+     "iopub.status.busy": "2022-04-25T18:03:43.616170Z",
+     "iopub.status.idle": "2022-04-25T18:03:54.114821Z",
+     "shell.execute_reply": "2022-04-25T18:03:54.114285Z",
+     "shell.execute_reply.started": "2022-04-25T18:03:43.616427Z"
```
jsignell, line 63: It seems slightly odd to me that this is all in here.

jnolis (approved):

PR #363, opened 2022-04-28
Plumber Example: The / endpoint redirects to the docs page for the app

ms4952:
added endpoint / , which redirects to the DOCS ( swagger UI ) of plumber

### DIFF
--- a/examples/r-api/README.md
+++ b/examples/r-api/README.md
@@ -22,7 +22,7 @@ install.packages("plumber")
 install.packages("dplyr")
 ```
 
-Now let us create another R script called `endpoints.R`. This script will contain the endpoints for our plumber API, which in this case will be `/` and `/predict`. Endpoint `/` redirects to interactive API documentation. Endpoint `/predict` calls a regression model and predics house price. 
+Now let us create another R script called `endpoints.R`. This script will contain the endpoints for our plumber API, which in this case will be `/` and `/predict`. The endpoint `/` redirects to the interactive API documentation. The endpoint `/predict` calls a regression model and predics house price. 
 
 In the following code in first line we have placed oxygen2-like comments for describing what this API will do. Now we will define the operation and path. Here we have used `GET` operation (to read data) and path is `/predict`. This endpoint will return the predicted price of a house.
 
--- a/examples/r-api/README.md
+++ b/examples/r-api/README.md
@@ -52,7 +52,7 @@ function(bedrooms,year,res){
 #* Redirect to Docs
 #* @get /
 function(res) {
-  res$status <- 303 # redirect
+  res$status <- 302 # redirect
   res$setHeader("Location", "./__docs__/")
 }
 ```
--- a/examples/r-api/README.md
+++ b/examples/r-api/README.md
@@ -22,11 +22,15 @@ install.packages("plumber")
 install.packages("dplyr")
 ```
 
-Now let us create another R script called `endpoints.R`. This script will contain the endpoints for our plumber API, which in this case will be a single endpoint that calls a regression model. In the following code in first line we have placed oxygen2-like comments for describing what this API will do. Now we will define the operation and path. Here we have used `GET` operation (to read data) and path is `/predict`. This endpoint will return the predicted price of a house.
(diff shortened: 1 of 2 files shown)


examples/r-api/README.md
```diff
@@ -45,6 +49,12 @@ function(bedrooms,year,res){
     return(list(error = "Please enter BedroomAbvGr between 0 and 8. Enter YearBuilt between 1872 and 2100."))
   }  
 }
+#* Redirect to Docs
+#* @get /
+function(res) {
+  res$status <- 303 # redirect
```
jsignell: How did you choose 303? I think 301 would be more normal (docs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections)

ms4952: I choose since it stands for redirecting to a different resource https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303 . But if 301 makes more sense here, I can replace with that ,

jsignell: I just checked with Alex and he says 302 is probably best.

ms4952: okie cool .. Will change in the other pr as well

jsignell (approved):
LGTM

examples/r-api/README.md
```diff
@@ -22,12 +22,16 @@ install.packages("plumber")
 install.packages("dplyr")
 ```
 
-Now let us create another R script called `endpoints.R`. This script will contain the endpoints for our plumber API, which in this case will be a single endpoint that calls a regression model. In the following code in first line we have placed oxygen2-like comments for describing what this API will do. Now we will define the operation and path. Here we have used `GET` operation (to read data) and path is `/predict`. This endpoint will return the predicted price of a house.
+Now let us create another R script called `endpoints.R`. This script will contain the endpoints for our plumber API, which in this case will be `/` and `/predict`. Endpoint `/` redirects to interactive API documentation. Endpoint `/predict` calls a regression model and predics house price.
```
jnolis:
```suggestion
Now let us create another R script called `endpoints.R`. This script will contain the endpoints for our plumber API, which in this case will be `/` and `/predict`. The endpoint `/` redirects to the interactive API documentation. The endpoint `/predict` calls a regression model and predics house price. 
```

examples/r-api/README.md
```diff
@@ -22,12 +22,16 @@ install.packages("plumber")
 install.packages("dplyr")
 ```
 
-Now let us create another R script called `endpoints.R`. This script will contain the endpoints for our plumber API, which in this case will be a single endpoint that calls a regression model. In the following code in first line we have placed oxygen2-like comments for describing what this API will do. Now we will define the operation and path. Here we have used `GET` operation (to read data) and path is `/predict`. This endpoint will return the predicted price of a house.
+Now let us create another R script called `endpoints.R`. This script will contain the endpoints for our plumber API, which in this case will be `/` and `/predict`. Endpoint `/` redirects to interactive API documentation. Endpoint `/predict` calls a regression model and predics house price. 
+
+In the following code in first line we have placed oxygen2-like comments for describing what this API will do. Now we will define the operation and path. Here we have used `GET` operation (to read data) and path is `/predict`. This endpoint will return the predicted price of a house.
```
jnolis, line 27:
```suggestion
In the following code in the first line we have placed oxygen2-like comments for describing what this API will do. Now we will define the operation and path. Here we have used `GET` operation and path is `/predict`. This endpoint will return the predicted price of a house.
```

examples/r-api/README.md
```diff
@@ -22,12 +22,16 @@ install.packages("plumber")
...
 
 This function is taking 3 arguments: `response`, `bedrooms` which represents number of bedrooms and `year` which represents year build. `bedrooms` and `year` are inputs from the client.
 
 Inside function we read houseprice data and perform linear regression to build a model. We then predict price of house for given inputs. We are returning HTTP response 400 if parameters passed are not in range of training data. If the parameters fall within valid range, API will return predicted house price.
 
+The path for second endpoint is `/`. We are redirecting this endpoint to `/__docs__/`, which is interactive API documentation. In this documentation UI you can enter the required values and trigger response.
```
jnolis, line 33:
```suggestion
The path for the second endpoint is `/`. We are redirecting this endpoint to `/__docs__/`, which is the interactive API documentation. In this documentation UI you can enter the required values and trigger a response.
```

jnolis (approved):
Minor grammar changes but otherwise good!

ms4952 (commented):
changes from feedback

PR #364, opened 2022-04-29
add credentials=none for anon access to AWS for julia dashboard example

nathanballou:
This adds antonymous credentials for the Julia dashboard example. Without this change, we would need to specify the AWS credentials on the example template.


jsignell (approved):
Makes sense to me!

PR #365, opened 2022-05-03
Prefect cloud flow runs create their own dask clusters

bhperry:
Noticed dask clusters being created on workspace while testing prefect example. Flow runs get their own dask clusters now, so there is no need to create one on the workspace.

### DIFF
--- a/examples/prefect/.saturn/saturn.json
+++ b/examples/prefect/.saturn/saturn.json
@@ -17,15 +17,6 @@
     "disk_space": "10Gi",
     "instance_type": "large"
   },
-  "dask_cluster": {
-    "num_workers": 3,
-    "worker": {
-      "instance_type": "large"
-    },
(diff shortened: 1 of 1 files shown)


jnolis:
Adding @ms4952 since she most recently redid it

ms4952:
LGTM,  I will change documentation and example to align with this.

jsignell (commented):
This example isn't entirely about prefect cloud flows though right? Is the dask cluster needed for the non-cloud examples?

bhperry:
> This example isn't entirely about prefect cloud flows though right? Is the dask cluster needed for the non-cloud examples?

It seems to be. All three notebooks here register flows with prefect-cloud

jsignell:
ah ok maybe I had that idea from an old version

jsignell (approved):

PR #368, opened 2022-05-05
Fast-API Example: The / endpoint redirects to the docs page for the app

ms4952:
added endpoint / , which redirects to the DOCS  of rest-api

### DIFF
--- a/examples/api/houseprice.py
+++ b/examples/api/houseprice.py
@@ -1,6 +1,7 @@
 from fastapi import FastAPI, HTTPException
 import pandas as pd
 from sklearn.linear_model import LinearRegression
+from starlette.responses import Response
 
 app = FastAPI()
 df = pd.read_csv(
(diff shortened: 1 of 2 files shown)


examples/api/houseprice.py
```diff
@@ -10,6 +11,11 @@
 lr.fit(df[["BedroomAbvGr", "YearBuilt"]], df["SalePrice"])
 
 
+@app.get("/")
+async def docs_redirect():
+    return Response("Opening the docs UI", status_code=307, headers={"location": "/docs"})
```
jsignell: Since we are explicitly setting the status let's use something more typical, I suggest 302

ms4952: @jsignell  should I change this to 301 as well . 307 is fortemporary redirect ( message that I was getting in logs)

ms4952: okie

jsignell (approved):

jnolis (approved):

PR #375, opened 2022-06-14
adding Julia workshop code

nathanballou:
Adds the Julia workshop code to the examples.


jnolis (approved):
My only suggestion is you might want to update the recipe schema to a later version. 1.06 is pretty old!

jsignell:
All the recipe schemas in examples have to use the same schema version. That is intended to make things easier to keep track of.

PR #380, opened 2022-06-28
opencv example

ms4952:
This is an opencv example. We are doing  image processing using opencv features. For this example (in recipe) I have installed opencv-python in extra packages, since the pytorch image is only for GPUs and we did not need that much computation for this example.


jnolis:
I think this example needs some fundamental changes--I slacked @ms4952 to discuss it

PR #407, opened 2025-11-27
Saturn Cloud Template: Data Engineering & Pipelines

GeoSegun:
These are templates:

- [x] nvidia-spark-rapids
- [x] stream-ingest


GeoSegun:
moved to #408

